PR #1, opened 2017-07-14
swagger spec for the filter api

Matt-Guest:
### What

A swagger spec to filter datasets

### How to review
review the endpoints provided by the spec

### Who can review
anyone in team a and b

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -1,6 +1,6 @@
 swagger: "2.0"
 info:
-  description: "An API used to customise a dataset. Once the job has been created it can then be updated with filters to apply to the dataset. Once the job is ready it is expected that the client will update the job state to start the processing. Once processing has been started the job will be placed in a locked condition as no amendments can be done."
+  description: "An API used to customise and submit queries (filters) against a given dataset. Once the filter job has been created it can then be updated with dimension-specific filters that apply to that dataset. Once all the required filters have been applied to the job, it is expected that the client will update the job state to indicate that processing can begin. Once processing has been started the job will be placed in a locked condition such that no further amendments can be done. Finally, the state will be marked as completed (or failed)."
   version: "1.0.0"
   title: "Filter API"
   license:
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -33,6 +33,13 @@ parameters:
     required: true
     description: "The requested state of the job"
     in: body
+  newJob:
+    name: newJob
+    schema:
+      $ref: '#/definitions/NewJob'
+    required: true
+    description: "A new job to create"
+    in: body
 paths:
   /job:
     post:
@@ -42,11 +49,13 @@ paths:
       description: "Create a job so that dimensions can be added to filter a dataset"
       produces:
       - "application/json"
+      parameters:
+      - $ref: '#/parameters/newJob'
       responses:
         200:
           description: "Job was created"
           schema:
-            $ref: '#/definitions/Job'
+            $ref: '#/definitions/JobInstance'
         500:
           description: "Failed to process the request due to an internal error"
   /jobs/{jobId}:
@@ -109,12 +118,31 @@ paths:
         500:
           description: "Failed to process the request due to an internal error"
 definitions:
-  Job:
+  JobInstance:
     type: object
     properties:
       id:
         type: string
         description: "An unique id for the filter job"
+  NewJob:
+    description: |
+      A model used to create a new filter job
+    type: object
+    required: ["dataset", "edition", "version"]
+    properties:
+      dataset:
+        type: string
+        description: "The type of dataset"
+      edition:
+        type: string
+        description: "The edtion of the dataset (This could be monthly or yearly version)"
+      version:
+        type: string
+        description: "The version of the dataset (This includes any amendments to the original dataset edition)"
+      filters:
+        type: array
+        items:
+          $ref: '#/definitions/Dimension'
   JobState:
     description: |
       A description of a job to generate a customised dataset
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -61,6 +61,8 @@ paths:
           description: "Job was created"
           schema:
             $ref: '#/definitions/JobState'
+        400:
+          description: "Invalid request body"
         500:
           $ref: '#/responses/InternalError'
   /filters/{filter_id}:
@@ -136,9 +138,13 @@ paths:
         200:
           description: "The dimension was created"
         400:
-          description: "The filter job has been locked as it has been `submitted` to be processed"
+          description: |
+            This error code could be one or more of:
+            * The filter job has been locked as it has been `submitted` to be processed
+            * Invalid request body
+            * Filter job was not found
         404:
-          $ref: '#/responses/FilterOrDimensionNotFound'
+          description: "Dimension name not found"
         500:
           $ref: '#/responses/InternalError'
     delete:
@@ -150,9 +156,12 @@ paths:
         200:
           description: "The dimension was removed"
         400:
-          description: "The filter job has been locked as it has been `submitted` to be processed"
+          description: |
+            This error code could be one or more of:
+            * The filter job has been locked as it has been `submitted` to be processed
+            * Filter job was not found
         404:
-          $ref: '#/responses/FilterOrDimensionNotFound'
+          description: "Dimension name was not found"
         500:
           $ref: '#/responses/InternalError'
     get:
@@ -166,9 +175,9 @@ paths:
           schema:
             $ref: '#/definitions/Dimension'
         400:
-          description: "The filter job has been locked as it has been `submitted` to be processed"
+          description: "Filter job was not found"
         404:
-          $ref: '#/responses/FilterOrDimensionNotFound'
+          description: "Dimension name was not found"
         500:
           $ref: '#/responses/InternalError'
   /filters/{filter_id}/dimensions/{name}/options/{value}:
@@ -187,7 +196,7 @@ paths:
         400:
           $ref: '#/responses/FilterOrDimensionNotFound'
         404:
-          description: "Filter and dimension exist but value was not found"
+          description: "Dimension value was not found"
         500:
           $ref: '#/responses/InternalError'
     post:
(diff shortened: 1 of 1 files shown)


swagger.yaml
```diff
@@ -0,0 +1,221 @@
...
+    properties:
+      dataset:
+        type: string
+        description: "The type of dataset"
+      edition:
+        type: string
+        description: "The edtion of the dataset (This could be monthly or yearly version)"
+      version:
+        type: string
+        description: "The version of the dataset (This includes any amendments to the original dataset edition)"
```
CarlHembrough: I am a bit confused about how dimensions are defined. I think what you call dimensions I think of as dimension options. For example, a dimension would be gender, with the options male and female. What we are adding here are dimension options.

swagger.yaml
```diff
@@ -0,0 +1,221 @@
...
+      responses:
+        200:
+          description: "Filter list was updated"
+        400:
+          description: "The job has been locked as it has been `submitted` to be processed"
+        404:
+          description: "Job not found"
+        500:
+          description: "Failed to process the request due to an internal error"
+  /jobs/{jobId}/state:
```
CarlHembrough: As state is a property of the job, it should be a PUT to /jobs/{jobId} specifying only the state in the body.

swagger.yaml
```diff
@@ -0,0 +1,221 @@
...
+      responses:
+        200:
+          description: "The job was found and a `JobState` is returned"
+          schema:
+            $ref: '#/definitions/JobState'
+        404:
+          description: "Job not found"
+        500:
+          description: "Failed to process the request due to an internal error"
+  /jobs/{jobId}/filter:
```
CarlHembrough: We will need a way to delete existing filters. This may open up a few considerations like how to reference existing filters. We can discuss this further when your back in next week.

swagger.yaml
```diff
@@ -0,0 +1,221 @@
...
+      A model used to create a new filter job
+    type: object
+    required: ["dataset", "edition", "version"]
+    properties:
+      dataset:
+        type: string
+        description: "The type of dataset"
+      edition:
+        type: string
+        description: "The edtion of the dataset (This could be monthly or yearly version)"
```
eldeal: Typo on edition, and again on 156

swagger.yaml
```diff
@@ -0,0 +1,333 @@
...
+- name: "Filter"
+  description: "Used to filter published datasets"
+schemes:
+- "http"
+parameters:
+  filter_id:
+    name: filter_id
+    type: string
+    required: true
+    description: "The unique  filter ID for customising a dataset"
```
CarlHembrough: Double space (!)

swagger.yaml
```diff
@@ -0,0 +1,333 @@
...
+            $ref: '#/definitions/JobState'
+        500:
+          $ref: '#/responses/InternalError'
+  /filters/{filter_id}:
+    parameters:
+      - $ref: '#/parameters/filter_id'
+    get:
+      tags:
+      - "Filter"
+      summary: "Get a description of a created job"
```
CarlHembrough: 'created filter' ?

swagger.yaml
```diff
@@ -0,0 +1,333 @@
...
+      description: "A jobId is required to access a customised dataset job. To create a jobId see the `Dataset API`"
```
CarlHembrough: This description needs updating. Still references jobId and mentions creating a job via the dataset API.

swagger.yaml
```diff
@@ -0,0 +1,333 @@
...
+          description: "The filter job has been locked as it has been `submitted` to be processed"
+        404:
+          $ref: '#/responses/FilterOrdimensionNotFound'
+        500:
+          $ref: '#/responses/InternalError'
+responses:
+  FilterJobNotFound:
+    description: "Filter Job not found"
+  FilterOrdimensionNotFound:
+    description: "The filter job has been locked as it has been `submitted` to be processed"
```
CarlHembrough: Is this the right message? I was expecting it to just be something like 'Dimension not found.' A number of the endpoints are returning the same message for 400 and 404 error?

swagger.yaml
```diff
@@ -0,0 +1,333 @@
...
+        400:
+          description: "The filter job has been locked as it has been `submitted` to be processed"
+        404:
+          $ref: '#/responses/FilterOrdimensionNotFound'
+        500:
+          $ref: '#/responses/InternalError'
+responses:
+  FilterJobNotFound:
+    description: "Filter Job not found"
+  FilterOrdimensionNotFound:
```
CarlHembrough: Dimension needs a capital letter

swagger.yaml
```diff
@@ -0,0 +1,333 @@
...
+    name: dimension_name
+    type: string
+    required: true
+    description: "The name of a dimension"
+    in: path
+  value:
+    name: value
+    type: string
+    required: true
+    description: "The a single value for a dimension"
```
CarlHembrough: Typo here - "The a..."

swagger.yaml
```diff
@@ -0,0 +1,333 @@
...
+    description: "Failed to process the request due to an internal error"
+definitions:
+  NewJob:
+    description: "A model used to create new filter jobs. Dimensions are optional"
+    allOf:
+    - $ref: '#/definitions/JobState'
+    - type: object
+      properties:
+        dimensions:
+          type: array
```
nshumoogum: As discussed this should have a readOnly set to false.

swagger.yaml
Matt-Guest: Thats a typo =(

swagger.yaml
```diff
@@ -0,0 +1,333 @@
...
+    get:
+      tags:
+      - "Filter"
+      summary: "Check if a value exists within a dimension"
+      description: "Check if a value exists within a dimension"
+      responses:
+        200:
+          description: "Value exists within the dimension"
+        400:
+          $ref: '#/responses/FilterOrdimensionNotFound'
```
nshumoogum: The reason for a 400 (bad request) here would be because the get request `filter_id` or `dimension_name` don't exist.

A 404 (not found) applies when the `filter_id` and `dimension_name` exist but the `value` does not exist.

swagger.yaml
```diff
@@ -0,0 +1,333 @@
...
+      tags:
+      - "Filter"
+      summary: "Get all values from a dimension which have been set"
+      description: "Get a list of all values which will be used to filter the dimension"
+      responses:
+        200:
+          description: "A list of all values for a dimension was returned"
+          schema:
+            $ref: '#/definitions/Dimension'
+        400:
```
nshumoogum: Would we respond with status 400 on a GET request for locking a submitted filter?

Also a 400 could be returned if the `filter_id` does not exist and a 404 if the `dimension_name` does not exist.

swagger.yaml
```diff
@@ -0,0 +1,334 @@
...
+    get:
+      tags:
+      - "Filter"
+      summary: "Check if a value exists within a dimension"
+      description: "Check if a value exists within a dimension"
+      responses:
+        200:
+          description: "Value exists within the dimension"
+        400:
+          $ref: '#/responses/FilterOrDimensionNotFound'
```
nshumoogum: The reason for a 400 (bad request) here would be because the get request filter_id or dimension_name don't exist.

A 404 (not found) applies when the filter_id and dimension_name exist but the value does not exist.

swagger.yaml
```diff
@@ -0,0 +1,334 @@
...
+          description: "A list of dimension URLs"
+          schema:
+            type: array
+            items:
+              type: string
+        404:
+          $ref: '#/responses/FilterJobNotFound'
+        500:
+          $ref: '#/responses/InternalError'
+  /filters/{filter_id}/dimensions/{dimension_name}:
```
nshumoogum: Do we want to keep consistent naming convention from the uri parameters and the fields received in the response body and sent in the request body. For instance `dimension_name` in the uri is equivalent to the `name` in the model, should the uri not be /filters/{filter_id}/dimensions/{name}

Matt-Guest: Good spot, this also makes the uri smaller and stops the word dimension being repeated 👍

swagger.yaml
```diff
@@ -0,0 +1,334 @@
+swagger: "2.0"
+info:
+  description: "An API used to customise a dataset. Once the job has been created it can then be updated with filters to apply to the dataset. Once the job is ready it is expected that the client will update the job state to start the processing. Once processing has been started the job will be placed in a locked condition as no amendments can be done."
```
gedge: I think this wording sounds better:

An API used to customise and submit queries (filters) against a given dataset. Once the filter job has been created it can then be updated with dimension-specific filters that apply to that dataset. Once all the required filters have been applied to the job, it is expected that the client will update the job state to indicate that processing can begin. Once processing has been started the job will be placed in a locked condition such that no further amendments can be done.  Finally, the state will be marked as completed (or failed).

swagger.yaml
```diff
@@ -0,0 +1,350 @@
...
+        Return a list of all dimensions which are going to be used to filter on
+      parameters:
+      - $ref: '#/parameters/filter_id'
+      responses:
+        200:
+          description: "A list of dimension URLs"
+          schema:
+            type: array
+            items:
+              type: string
```
mattrout92: Small comment here, but would it be worth wrapping this array of string with a links descriptor?

For example

```json
"links" : [
    "path1",
    "path2",
    "path3"
]
```

just makes it really clear to the user that these are links and not just a random list of strings

swagger.yaml
Matt-Guest: Thanks for that @gedge , that a better description for the API. I Updated the spec with it

swagger.yaml
Matt-Guest: @mattrout92 Thanks for the feedback. We updated the returned type and the model.

swagger.yaml
Matt-Guest: Done

swagger.yaml
Matt-Guest: Done

nshumoogum:
Looks even better to me 👍

PR #4, opened 2017-08-03
Add new endpoint to update filter job status

nshumoogum:
### What

Additional endpoint to update filter job (specifically the status),  PUT request on `/filters/<filter_id>`

1) Includes test coverage
2) If state is updated to submitted then message is written to kafka
topic.
3) Test coverage for producing messages using mocks package
4) Extend config package to include new configurable variables
5) Update README.md

### How to review

Check tests reach a high level of code coverage and check tests run successfully. Check code meets coding standards.

### Who can review

Team B


postgres/postgres.go
```diff
@@ -56,7 +56,44 @@ func (ds Datastore) AddFilter(host string, newFilter *models.Filter) (models.Fil
...
+	}
+
+	stmt, err := tx.Prepare(updateStatement)
+	if err != nil {
+		return err
+	}
+
+	_, err = stmt.Exec(filter.FilterID)
+	if err != nil {
+		return err
```
Matt-Guest: For all the error cases here, would you need to rollback the transaction?

postgres/postgres.go
```diff
@@ -79,10 +116,68 @@ func (ds Datastore) addDimensionValues(tx *sql.Tx, filterID string, dimension mo
...
+		return filterJob, convertSQLError(err)
+	}
+
+	state := filter.String
+
+	filterJob.FilterID = filterID
+	filterJob.State = state
+
+	if state == "submitted" {
+		err := fmt.Errorf("Forbidden")
```
Matt-Guest: Should this be "errors.New()"?

postgres/postgres.go
```diff
@@ -79,10 +116,68 @@ func (ds Datastore) addDimensionValues(tx *sql.Tx, filterID string, dimension mo
...
+	statement := "UPDATE Filters SET"
+	var hasUpdate bool
+
+	if filter.State != "" {
+		hasUpdate = true
+		statement = statement + " state = '" + filter.State + "',"
+	}
+
+	if !hasUpdate {
+		err := fmt.Errorf("Bad request")
```
Matt-Guest: Should this be "errors.New()"?

postgres/postgres.go
nshumoogum: So no as I haven't been able to change anything against the database.

The statement previously only retrieves the data to check the state before doing an update to it. So to stop another update or delete to happen against this data (filter job) I have placed both sql statements in a transaction to lock the row, this will prevent any changes occurring between the two queries/statements.

PR #7, opened 2017-08-08
Include new GET request endpoints

nshumoogum:
### What

* New endpoints to include getting list of dimension urls, list of
 dimension option urls and check if either dimension or option exists
* Extend tests to cover new endpoints
* Handle new response messages when status response is not found
* Update field name `value` to `option` in postgres table and hence
 update script

### How to review

### Who can review

Team B

### DIFF
--- a/postgres/postgres.go
+++ b/postgres/postgres.go
@@ -11,28 +11,34 @@ import (
 
 // Datastore represents a structure to hold SQL statements to be used to gather information or insert about filters and dimensions
 type Datastore struct {
-	db                 *sql.DB
-	addFilter          *sql.Stmt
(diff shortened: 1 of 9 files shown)


postgres/postgres.go
```diff
@@ -233,9 +376,18 @@ func prepare(sql string, db *sql.DB) (*sql.Stmt, error) {
 	return statement, nil
 }
 
-func convertSQLError(err error) error {
+func convertSQLError(err error, typ string) error {
```
nshumoogum: type is a reserved word, hence typ

Matt-Guest: Ahh I forgot about that.

PR #13, opened 2017-08-15
 Extend PUT request to handle authentication header

nshumoogum:
### What

Extend PUT request to handle authentication header; this allows backend services to update the filter job when the state is set to `submitted`.

Also includes: 

* Additional logic on endpoints to update filter job to prevent users updating a job if it is already completed
* Update swagger spec
* Correctly returned error messages to match swagger spec

### How to review

### Who can review

Team B

### DIFF
--- a/api/filters.go
+++ b/api/filters.go
@@ -284,7 +284,7 @@ func (api *FilterAPI) updateFilterJob(w http.ResponseWriter, r *http.Request) {
 
 	filter.FilterID = filterID
 
-	err = api.dataStore.UpdateFilter(isAuthenticated, filterID, filter)
+	err = api.dataStore.UpdateFilter(isAuthenticated, filter)
 	if err != nil {
 		log.Error(err, log.Data{"filter": filter, "filter_job_id": filterID})
 		setErrorCode(w, err)
--- a/api/filters.go
+++ b/api/filters.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"encoding/json"
+	"errors"
 	"net/http"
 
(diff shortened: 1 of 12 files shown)


api/filters.go
```diff
@@ -257,16 +262,29 @@ func (api *FilterAPI) addFilterJobDimensionOption(w http.ResponseWriter, r *http
...
+			log.Error(err, log.Data{"filter_job_id": filterID})
+			setErrorCode(w, err)
+			return
+		}
+	}
+
 	filter.FilterID = filterID
 
-	err = api.dataStore.UpdateFilter(filterID, filter)
+	err = api.dataStore.UpdateFilter(isAuthenticated, filterID, filter)
```
eldeal: Why pass the ID through separately to the filter, when the ID is contained in the filter anyway?

postgres/postgres.go
```diff
@@ -545,10 +548,12 @@ func getFilterJobState(tx *sql.Tx, ds Datastore, filterID string) (models.Filter
...
 
-	if state == submittedState {
-		err := errors.New("Forbidden")
-		log.ErrorC("update to filter job forbidden, job already submitted", err, log.Data{"filter_job_id": filterID})
-		return filterJob, err
+	if !isAuthenticated {
+		if state == submittedState || state == completedState {
+			err := errors.New("Forbidden")
+			log.ErrorC("update to filter job forbidden, job already submitted", err, log.Data{"filter_job_id": filterID})
+			return filterJob, err
```
eldeal: `return nil, err` - no point returning a filterjob that is gonna get ignored because the error returns

PR #14, opened 2017-08-16
Add nomad file.

CarlHembrough:
### What

Add nomad file.


dp-filter-api.nomad
```diff
@@ -0,0 +1,100 @@
+job "dp-filter-api" {
+  datacenters = ["eu-west-1"]
+  region      = "eu"
+  type        = "service"
+
+  // Make sure that this API is only ran on the publishing nodes
+  constraint {
+    attribute = "${node.class}"
```
eldeal: Does this need to only run on publishing? The group below says web which seems accurate

CarlHembrough: Fixed this after speaking with Lloyd - for services that run on both web and publishing the constraint needs to be inside the group for both web and publishing groups.

PR #15, opened 2017-08-17
Authenticated users allowed to update Downloads table vi 

nshumoogum:
### What

Update PUT endpoint `/filters/<filter_job_id>` to allow Downloads table to be updated by authenticated users. 

If authentication fails return error status 401 (unauthenticated), if auth token is not set in header of request and the filter job has been submitted then error status returned is 403 (forbidden) otherwise returns a 400 (bad request) when a user trys to update downloads table.

Also includes:

* Fix idle connections, these were previously not being closed on calling external Query method.

### How to review

### Who can review

Team B

### DIFF
--- a/postgres/postgres.go
+++ b/postgres/postgres.go
@@ -607,19 +607,19 @@ func (ds Datastore) removeDimension(tx *sql.Tx, dimensionObject *models.AddDimen
 }
 
 func getFilterJobState(tx *sql.Tx, ds Datastore, filterID string, isAuthenticated bool) (*models.Filter, error) {
-	var filterJob *models.Filter
 	row := tx.Stmt(ds.getFilterState).QueryRow(filterID)
 
 	var filter sql.NullString
-
 	if err := row.Scan(&filter); err != nil {
 		return nil, convertError(err, "")
 	}
 
 	state := filter.String
 
-	filterJob.FilterID = filterID
-	filterJob.State = state
+	filterJob := &models.Filter{
+		FilterID: filterID,
+		State:    state,
+	}
 
 	if !isAuthenticated {
 		if state == submittedState || state == completedState {
--- a/postgres/postgres.go
+++ b/postgres/postgres.go
@@ -632,13 +632,6 @@ func getFilterJobState(tx *sql.Tx, ds Datastore, filterID string, isAuthenticate
 	return filterJob, nil
 }
 
-type statements struct {
-	filterStatement       string
-	downloadCSVStatement  string
-	downloadJSONStatement string
-	downloadXLSStatement  string
-}
-
 func prepare(sql string, db *sql.DB) (*sql.Stmt, error) {
 	statement, err := db.Prepare(sql)
 	if err != nil {
--- a/postgres/postgres.go
+++ b/postgres/postgres.go
@@ -3,15 +3,19 @@ package postgres
 import (
 	"database/sql"
 	"errors"
-	"strings"
 
 	"github.com/ONSdigital/dp-filter-api/models"
 	"github.com/ONSdigital/go-ns/log"
+	"github.com/lib/pq"
 )
 
 const (
 	submittedState = "submitted"
 	completedState = "completed"
+
+	CSV  = "csv"
+	JSON = "json"
+	XLS  = "xls"
 )
 
 // Datastore represents a structure to hold SQL statements to be used to gather information or insert about filters and dimensions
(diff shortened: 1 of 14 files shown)


postgres/postgres.go
```diff
@@ -465,26 +493,71 @@ func (ds Datastore) UpdateFilter(isAuthenticated bool, filter *models.Filter) er
 		return err
 	}
 
-	updateStatement, err := updateStatement(filter)
-	if err != nil {
-		return err
+	rowsAffected := int64(0)
+
+	if filter.State != "" {
```
CarlHembrough: Why would we not update the filter state if its currently empty?

postgres/postgres.go
```diff
@@ -465,26 +493,71 @@ func (ds Datastore) UpdateFilter(isAuthenticated bool, filter *models.Filter) er
...
+			}
+			return err
+		}
 	}
 
-	stmt, err := tx.Prepare(updateStatement)
-	if err != nil {
-		return err
+	if !isAuthenticated && rowsAffected == 0 {
+		return errors.New("Bad request")
```
CarlHembrough: Should this return a specific error allowing a 401 / 403 to be returned?

postgres/postgres.go
```diff
@@ -465,26 +493,71 @@ func (ds Datastore) UpdateFilter(isAuthenticated bool, filter *models.Filter) er
...
 	}
 
-	_, err = stmt.Exec(filter.FilterID)
-	if err != nil {
-		return err
+	csv := filter.Downloads.CSV
+	json := filter.Downloads.JSON
+	xls := filter.Downloads.XLS
+
+	if isAuthenticated {
```
CarlHembrough: You may not need to check isAuthenticated here. If the request is not authenticated the method would have already returned from the previous check?

postgres/postgres.go
```diff
@@ -465,26 +493,71 @@ func (ds Datastore) UpdateFilter(isAuthenticated bool, filter *models.Filter) er
 		return err
 	}
 
-	updateStatement, err := updateStatement(filter)
-	if err != nil {
-		return err
+	rowsAffected := int64(0)
+
+	if filter.State != "" {
+		results, err := tx.Stmt(ds.updateFilterState).Exec(filter.FilterID, filter.State)
```
CarlHembrough: Can we test that the expected query is executed in a transaction if the filter.State != ""

postgres/postgres.go
```diff
@@ -465,26 +493,71 @@ func (ds Datastore) UpdateFilter(isAuthenticated bool, filter *models.Filter) er
...
+			if err != nil {
+				if err = tx.Rollback(); err != nil {
+					return err
+				}
+				return err
+			}
+		}
+
+		if xls.Size != "" {
+			_, err := tx.Stmt(ds.upsertDownloadURL).Exec(filter.FilterID, xls.Size, XLS, xls.URL)
```
CarlHembrough: Can we test that the expected query is executed in a transaction if the size != ""?

nshumoogum:
In response to the above comment, authentication is handled in filters.go in the api package but this only returns 401 or code continues (this was done in a separate PR).

403 is returned once the service has checked the database to see if the state of filter job is `submitted` or `completed` otherwise continue (this is handled in postgres.go in the postgres package).

postgres/postgres.go
nshumoogum: Possibly 403, but external user might send in a bad json body full of variables we don't know about which would require a 400.

Solution might be to add validation in model to check the body sent into this endpoint and then this can be changed to 403.

postgres/postgres.go
nshumoogum: Correct

postgres/postgres_test.go
```diff
@@ -6,50 +6,408 @@ import (
...
+			WillReturnResult(sqlmock.NewResult(1, 1))
+
+		dimensionOptionObject := &models.AddDimensionOption{
+			FilterID: "123",
+			Name:     "age",
+			Option:   "12",
+		}
+
+		err = ds.AddFilterDimensionOption(dimensionOptionObject)
+		So(err, ShouldBeNil)
```
CarlHembrough: So(mock.ExpectationsWereMet(), ShouldBeNil) ?

postgres/postgres_test.go
```diff
@@ -6,50 +6,408 @@ import (
...
+		dimension2 := models.Dimension{
+			DimensionURL: "/filters/123/dimensions/time",
+			Name:         "time",
+		}
+
+		expectedDimensions = append(expectedDimensions, dimension1, dimension2)
+
+		dimensions, err := ds.GetFilterDimensions("123")
+		So(err, ShouldBeNil)
+		So(dimensions, ShouldResemble, expectedDimensions)
```
CarlHembrough: So(mock.ExpectationsWereMet(), ShouldBeNil) ?

PR #18, opened 2017-09-19
Update service to gracefully shutdown

nshumoogum:
### What

Service gracefully shuts down when a terminal or interrupt signal is called or a critical error is returned from http server or from the kafka producer. Once the graceful shutdown is called any incoming requests are blocked but current processes are able to complete within the (configurable) `SHUTDOWN_TIMEOUT` .

### How to review

Make sure service shuts down gracefully as described above and connections to postgres are released.

To check this works requires adding in `time.Sleep()` in the endpoint you are calling and manually shutting down kafka or sending a signal to interrupt or terminate service. Make sure processes finish as long as they complete within the default time (5 seconds), once all processes complete, shutdown straight away (dont hang around waiting for the 5 seconds to complete) . Check new requests fail once service starts to shutdown and if any processes that take longer than the `SHUTDOWN_TIMEOUT` to complete, then they end up not completing and shutdown of service occurs.

To check postgres connections not sitting in idle, run the following whilst doing the above tests:
`psql -U dp FilterJobs -f scripts/CheckPQActivity.sql`

### Who can review

Team B


api/api.go
```diff
@@ -1,10 +1,23 @@
...
 	"github.com/gorilla/mux"
 )
 
+const (
+	gracefulShutdownMsg = "graceful shutdown of http server complete"
+)
+
+var httpServer *server.Server
+var serverErrors chan error
+var once sync.Once
```
daiLlew: is this being used at all?

api/api.go
```diff
@@ -20,10 +33,29 @@ type FilterAPI struct {
 }
 
 // CreateFilterAPI manages all the routes configured to API
-func CreateFilterAPI(secretKey string, host string, router *mux.Router, dataStore DataStore, jobQueue JobQueue) *FilterAPI {
-	router.Path("/healthcheck").Methods("GET").HandlerFunc(healthCheck)
+func CreateFilterAPI(secretKey, host, bindAddr string, datastore DataStore, jobQueue JobQueue, errorChan chan error) {
+	router := mux.NewRouter()
+	_ = routes(secretKey, host, router, datastore, jobQueue)
```
daiLlew: DO you need the assignment at all?

main.go
```diff
@@ -1,24 +1,29 @@
...
 	"github.com/ONSdigital/go-ns/log"
-	"github.com/ONSdigital/go-ns/server"
-	"github.com/gorilla/mux"
 	_ "github.com/lib/pq"
 )
 
 func main() {
 	log.Namespace = "dp-filter-api"
 
+	signals := make(chan os.Signal)
```
daiLlew: is it worth giving this a buffer of 1? I know its unlikely but if a signal is sent before exec reaches the select block it will block the app. (Sorry I am being picky)

main.go
```diff
@@ -50,19 +55,40 @@ func main() {
 	}
 
 	jobQueue := filterJobQueue.CreateJobQueue(producer.Output())
-	router := mux.NewRouter()
 
-	s := server.New(cfg.BindAddr, router)
+	apiErrors := make(chan error)
```
daiLlew: Same here -  if the creating the filter api sends a error to this chan before the select block is reached it will block.

api/api.go
nshumoogum: You are correct!

api/api.go
```diff
@@ -1,10 +1,23 @@
...
+	"sync"
+
 	"github.com/ONSdigital/dp-filter-api/models"
+	"github.com/ONSdigital/go-ns/log"
+	"github.com/ONSdigital/go-ns/server"
 	"github.com/gorilla/mux"
 )
 
+const (
+	gracefulShutdownMsg = "graceful shutdown of http server complete"
```
CarlHembrough: Consider inlining this as its only used once

api/api.go
```diff
@@ -37,3 +69,8 @@ func CreateFilterAPI(secretKey string, host string, router *mux.Router, dataStor
 	api.router.HandleFunc("/filters/{filter_job_id}/dimensions/{name}/options/{option}", api.removeFilterJobDimensionOption).Methods("DELETE")
 	return &api
 }
+
+func Close(ctx context.Context) {
+	httpServer.Shutdown(ctx)
```
CarlHembrough: Check if an error is returned here and return it to be logged.

config/config.go
```diff
@@ -23,14 +26,17 @@ func Get() (*Config, error) {
...
 		FilterJobSubmittedTopic: "filter-job-submitted",
-		Host:                    "http://localhost:22100",
-		KafkaMaxBytes:           "2000000",
-		PostgresURL:             "user=dp dbname=FilterJobs sslmode=disable",
-		SecretKey:               "FD0108EA-825D-411C-9B1D-41EF7727F465",
+		Host:            "http://localhost:22100",
+		KafkaMaxBytes:   "2000000",
+		PostgresURL:     "user=dp dbname=FilterJobs sslmode=disable",
+		SecretKey:       "FD0108EA-825D-411C-9B1D-41EF7727F465",
+		ShutdownTimeout: defaultTimeout,
```
CarlHembrough: Not sure why your assigning a variable for this. Can you just inline `ShutdownTimeout: 5 * time.Second`

config/config_test.go
```diff
@@ -24,6 +25,7 @@ func TestSpec(t *testing.T) {
 				So(cfg.FilterJobSubmittedTopic, ShouldEqual, "filter-job-submitted")
 				So(cfg.KafkaMaxBytes, ShouldEqual, "2000000")
 				So(cfg.SecretKey, ShouldEqual, "FD0108EA-825D-411C-9B1D-41EF7727F465")
+				So(cfg.ShutdownTimeout, ShouldEqual, time.Duration(5*time.Second))
```
CarlHembrough: So(cfg.ShutdownTimeout, ShouldEqual, 5*time.Second)

main.go
```diff
@@ -50,19 +55,40 @@ func main() {
...
-		"bind_address": cfg.BindAddr,
-	})
+	api.CreateFilterAPI(cfg.SecretKey, cfg.Host, cfg.BindAddr, dataStore, &jobQueue, apiErrors)
 
-	_ = api.CreateFilterAPI(cfg.SecretKey, cfg.Host, router, dataStore, &jobQueue)
+	// Gracefully shutdown the application closing any open resources.
+	gracefulShutdown := func() {
+		log.Info(fmt.Sprintf("Shutdown with timeout: %s", cfg.ShutdownTimeout), nil)
+		ctx, cancel := context.WithTimeout(context.Background(), cfg.ShutdownTimeout)
+		defer cancel()
```
CarlHembrough: Not sure if a deferred call to cancel with actually be called because of os.Exit?

config/config.go
```diff
@@ -27,10 +30,11 @@ func Get() (*Config, error) {
 		BindAddr:                ":22100",
 		Brokers:                 []string{"localhost:9092"},
 		FilterJobSubmittedTopic: "filter-job-submitted",
-		Host:                    "http://localhost:22100",
-		KafkaMaxBytes:           "2000000",
-		PostgresURL:             "user=dp dbname=FilterJobs sslmode=disable",
-		SecretKey:               "FD0108EA-825D-411C-9B1D-41EF7727F465",
+		Host:            "http://localhost:22100",
```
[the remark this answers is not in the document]
nshumoogum: It does, I will move `Host:  "http://localhost:22100",` to the top of list of values, this will weirdly solve the alignment.

PR #20, opened 2017-10-03
Feature/instance incl version

gedge:
### What

when creating a job, get version of instance/dataset from dataset API and put that into filter doc

### Synchronise with other repos

- same branch needs merging at the same time https://github.com/ONSdigital/dp-dataset-api/pull/35
-  also, https://github.com/ONSdigital/dp-dataset-exporter/pull/7


config/config.go
```diff
@@ -35,6 +37,7 @@ func Get() (*Config, error) {
 		PostgresURL:             "user=dp dbname=FilterJobs sslmode=disable",
 		SecretKey:               "FD0108EA-825D-411C-9B1D-41EF7727F465",
 		ShutdownTimeout:         5 * time.Second,
+		DatasetAPIURL:           "http://localhost:22000",
```
nshumoogum: Do we want a default `DatasetAPIAuthToken` set? I understand that if it changes in the dataset API we would have to change it here to.

I guess another argument is it forces the user of the service to set their own auth tokens instead of relying on insecure defaults.

config/config.go
```diff
@@ -16,6 +16,8 @@ type Config struct {
 	PostgresURL             string        `envconfig:"POSTGRES_URL"`
 	SecretKey               string        `envconfig:"SECRET_KEY"`
 	ShutdownTimeout         time.Duration `envconfig:"SHUTDOWN_TIMEOUT"`
+	DatasetAPIURL           string        `envconfig:"DATASET_API_URL"`
```
nshumoogum: The README.md needs to be updated with these new configurations (line 19 and 20)

nshumoogum:
CI (build.yml and unit.yml) needs updating to use go version 1.9.0 otherwise it will fail deployment, (tests fail in version 1.8.3)

Makefile
```diff
@@ -9,9 +10,9 @@ export GOARCH?=$(shell go env GOARCH)
 
 build:
 	@mkdir -p $(BUILD_ARCH)/$(BIN_DIR)
-	go build -o $(BUILD_ARCH)/$(BIN_DIR)/dp-filter-api main.go
+	go build -o $(BUILD_ARCH)/$(BIN_DIR)/dp-filter-api cmd/$(MAIN)/main.go
```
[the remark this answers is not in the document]
gedge: *this is consistent with other ~CMD~DP services* - I can't remember if it was the golang training but it was recommended

config/config.go
gedge: we don't want a default set, for security reasons (github, etc), but we have decided to have one for now (dev convenience) - we will remove later - I have added the default

config/config.go
gedge: done

PR #27, opened 2017-10-16
GET, POST and DELETE for dimensions options

Matt-Guest:
### What

Added GET, POST and DELETE for dimension options

### How to review

review the code and test out code using the swagger spec

### Who can review

anyone

### DIFF
--- a/mongo/filterstore.go
+++ b/mongo/filterstore.go
@@ -14,10 +14,11 @@ const FiltersCollection = "filters"
 
 const Submitted = "submitted"
 
-var NotFound = errors.New("not found")
+var NotFound = errors.New("Not found")
 var NotAuthorised = errors.New("Not authorised")
 var DimensionNotFound = errors.New("Dimension not found")
 var Forbidden = errors.New("Forbidden")
+var OptionNot = errors.New("Option not found")
 
 // FilterStore containing all filter jobs stored in mongodb
 type FilterStore struct {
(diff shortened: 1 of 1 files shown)


mongo/filterstore.go
```diff
@@ -170,22 +171,105 @@ func (s *FilterStore) RemoveFilterDimension(filterID string, name string) error
...
+	dimensionSelect := bson.M{"dimensions": 1}
+	var result models.Filter
+	err := session.DB(Database).C(FiltersCollection).Find(queryFilter).Select(dimensionSelect).One(&result)
+	if err != nil {
+		if err == mgo.ErrNotFound {
+			return nil, NotFound
+		}
+		return nil, err
+	}
+	err = session.DB(Database).C(FiltersCollection).Find(queryDimension).Select(dimensionSelect).One(&result)
```
nshumoogum: Do we need to make another request to mongo? We have the result of the filter job which contains the dimensions and options?

Instead we can check for the dimension by ranging over array of dimensions, this would be faster than making a call to the database again with a different query as well as opening us up to a change to the document between steps (even if it is highly unlikely).

mongo/filterstore.go
```diff
@@ -170,22 +171,105 @@ func (s *FilterStore) RemoveFilterDimension(filterID string, name string) error
...
+	dimensionSelect := bson.M{"dimensions": 1}
+	var result models.Filter
+	err := session.DB(Database).C(FiltersCollection).Find(queryFilter).Select(dimensionSelect).One(&result)
+	if err != nil {
+		if err == mgo.ErrNotFound {
+			return NotFound
+		}
+		return err
+	}
+	err = session.DB(Database).C(FiltersCollection).Find(queryDimension).Select(dimensionSelect).One(&result)
```
nshumoogum: Next two calls to mongo are unnecessary for the same reasons as above.

PR #34, opened 2017-10-27
Replace dimension in list rather than adding to list

eldeal:
### What

Ensure we're not adding the same dimension multiple times by replacing it in the list it lives in and setting that list rather than adding to it.

### How to review

Ensure API requests look sane and tests pass
Using filter journey on the website, try adding all time values and returning to the filter hub page - you should only see 1 time dimension with the number of options filled in.

### Who can review

@mattrout92 & @nshumoogum or @Matt-Guest

### DIFF
--- a/mongo/filterstore.go
+++ b/mongo/filterstore.go
@@ -186,8 +186,8 @@ func (s *FilterStore) AddFilterDimension(dimension *models.AddDimension) error {
 	defer session.Close()
 
 	list, err := s.GetFilterDimensions(dimension.FilterID)
-	if err != nil {
-		return errFilterOrDimensionNotFound
+	if err != nil && err != errNotFound {
+		return errBadRequest
 	}
 
 	for i, item := range list {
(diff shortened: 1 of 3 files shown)


mongo/filterstore.go
```diff
@@ -185,10 +185,32 @@ func (s *FilterStore) AddFilterDimension(dimension *models.AddDimension) error {
...
-	d := models.Dimension{Name: dimension.Name, Options: dimension.Options, DimensionURL: url}
-	update := bson.M{"$addToSet": bson.M{"dimensions": d}}
+	d := models.Dimension{Name: dimension.Name, Options: dimension.Options, URL: url}
+
+	var found bool
+	var index int
+	for i, item := range list {
+		if item.Name == d.Name {
+			found = true
+			index = i
```
nshumoogum: You can replace dimension here and then there will be no need for the variable index.

Something like this:
```
for i, item := range list {
  if item.Name == d.Name {
    list[i] = d
    found = true
    break
  }
}

if !found {
  list = append(list, d)
}
```

PR #35, opened 2017-10-30
Update swagger specs to to separate filters from job

nshumoogum:
### What

Filter job does not exist anymore instead there is a filter blueprint
which a user can update as many times as they like even after they
submit the resource. Once submitted a filter output resource is created
and cannot be updated by external users. The filter output is then
updated by backend process once the downloads are available. In the
meantime the filter blueprint can be used to create another filter
output.

### How to review

Check specification makes sense, all error status's are accounted for.

### Who can review

Team A or B

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -1,6 +1,6 @@
 swagger: "2.0"
 info:
-  description: "An API used to customise and submit queries (filters) against a given dataset. Once the filter job has been created it can then be updated with dimension-specific filters that apply to that dataset. Once all the required filters have been applied to the job, it is expected that the client will update the job state to indicate that processing can begin. Once processing has been started the job will be placed in a locked condition such that no further amendments can be done. Finally, the state will be marked as completed (or failed)."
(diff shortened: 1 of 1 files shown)


swagger.yaml
```diff
@@ -1,24 +1,32 @@
...
+  description: "An API used to customise and submit queries (filters) against a given dataset. Once the filter blueprint has been created it can then be updated with dimension-specific filters that apply to that dataset. Once all the required filters have been applied to the blueprint, it is expected that the client will send a Put request with query parameter `submitted` set to `true` to indicate that processing can begin; this request will result in a filter output resource to be created, copying the data from the blueprint and setting a status of submitted. The response for this request will give a link to the created filter output document. Once processing has been started the filter output resource will be placed in a locked condition such that onlu authorised requests can update the resource. Finally, the state will be marked as completed with a complete list of downloads (or failed)."
```
Matt-Guest: There a typo here.  `placed in a locked condition such that onlu authorised requests`

Matt-Guest: Also i'm if wondering if a blueprint shouldn't have a state. The reason being `completed` will never happen, once created and submitted if I go back and edit the blueprint will it always be `submitted`?

swagger.yaml
```diff
@@ -261,24 +274,72 @@ paths:
...
+      responses:
+        200:
+          description: "The filter output was found and document is returned"
+          schema:
+            $ref: '#/definitions/FilterOutputResponse'
+        404:
+           $ref: '#/responses/FilterOutputNotFound'
+        500:
+          $ref: '#/responses/InternalError'
+    put:
```
Matt-Guest: Can you add a `security` tag for this endpoint

swagger.yaml
nshumoogum: Yes I thought about this before and you are right the filter blueprint wont have a state; the filter output will still have a state but it will go from created to completed (there wont be a submitted state as this is just the creation of this resource).

PR #40, opened 2017-11-20
Feature/preview download

Matt-Guest:
### What

Preview a filtered job.

### How to review

review the code, run manual tests and unit tests

### Who can review
anyone


api/filters.go
```diff
@@ -530,6 +530,48 @@ func (api *FilterAPI) removeFilterBlueprintDimensionOption(w http.ResponseWriter
...
+	setJSONContentType(w)
+	w.WriteHeader(http.StatusOK)
+	_, err = w.Write(bytes)
+	if err != nil {
+		log.Error(err, log.Data{"filter_id": filterID})
+		setErrorCode(w, err)
+		return
+	}
+
+	log.Info("preview filter blueprint", log.Data{"filter_blueprint_id": filterID, "dimensions": filterBluePrint.Dimensions})
```
CarlHembrough: Is it worth moving this up near the start of the function so that its available when errors occur?

cmd/dp-filter-api/main.go
```diff
@@ -51,12 +60,13 @@ func main() {
 
 	client := rchttp.DefaultClient
 	datasetAPI := dataset.NewDatasetAPI(client, cfg.DatasetAPIURL, cfg.DatasetAPIAuthToken)
-
+	observationStore := observation.NewStore(neo4jDriver)
+	previewDatasets := preview.PreviewDatasetStore{Store: observationStore, Limit: 20}
```
CarlHembrough: Make the default limit configurable?

cmd/dp-filter-api/main.go
```diff
@@ -51,12 +60,13 @@ func main() {
 
 	client := rchttp.DefaultClient
 	datasetAPI := dataset.NewDatasetAPI(client, cfg.DatasetAPIURL, cfg.DatasetAPIAuthToken)
-
+	observationStore := observation.NewStore(neo4jDriver)
```
CarlHembrough: Should the API / PreviewDatasetStore maintain a connection pool to Neo4j? It looks like its using a single connection ATM

preview/preview.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+	var results FilterPreview
+	row, err := rows.Read()
+	if err != nil {
+		return nil, err
+	}
+	headers := strings.Split(strings.TrimSpace(row), ",")
```
CarlHembrough: Consider cells that have escaped comma's in them. It may be worth seeing if the CSV package can be used to parse / split the row. In the dataset-exporter I created an io.reader implementation for csv rows. Could maybe use a CSV.reader alongside that? (may well be overkill !)

preview/preview.go
```diff
@@ -0,0 +1,65 @@
...
+	results.Headers = headers
+	results.NumberOfColumns = len(headers)
+	for true {
```
CarlHembrough: I don't think 'true' is required here

api/filters.go
eldeal: Should we even be using the blueprint or the output? Or both?  When they're saving an active blueprint, we'll want to run the preview to ensure they still have data - but for the frontend preview screen this should relate to the output, so that changes to the blueprint afterwards don't cause previews to change for outputs which already exist

cmd/dp-filter-api/main.go
Matt-Guest: The limit been removed and replaced with a HTTP query

cmd/dp-filter-api/main.go
Matt-Guest: Done, I made another PR to address this https://github.com/ONSdigital/dp-filter/pull/3

preview/preview.go
Matt-Guest: I ended up using the csv reader as this breaks when a CSV cell has a comma within quotes.

preview/preview.go
Matt-Guest: done

api/filters.go
```diff
@@ -530,6 +531,58 @@ func (api *FilterAPI) removeFilterBlueprintDimensionOption(w http.ResponseWriter
 	log.Info("delete filtered blueprint", log.Data{"filter_blueprint_id": filterID, "dimension": name})
 }
 
+func (api *FilterAPI) getFilterOutputPreview(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	filterID := vars["filter_output_id"]
+	requestedLimit := r.URL.Query().Get("limit")
+	var limit int64 = 20
+	if requestedLimit != "" {
+		limit, err := strconv.ParseInt(requestedLimit, 10, 64)
```
CarlHembrough: Consider making the limit variable type `int`, and using `strconv.Atoi` to convert it

api/filters.go
```diff
@@ -530,6 +531,58 @@ func (api *FilterAPI) removeFilterBlueprintDimensionOption(w http.ResponseWriter
 	log.Info("delete filtered blueprint", log.Data{"filter_blueprint_id": filterID, "dimension": name})
 }
 
+func (api *FilterAPI) getFilterOutputPreview(w http.ResponseWriter, r *http.Request) {
```
CarlHembrough: Consider adding test coverage to this function - ensuring that the expected limit is passed down when specified and the default is applied when its not.

preview/preview.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		return nil, err
+	}
+	headers := row
+	results.Headers = headers
+	// Replace the V4 header withe values
+	results.Headers[0] = "Values"
+	results.NumberOfColumns = len(headers)
+	for {
+		row, err = csvReader.Read()
```
CarlHembrough: Consider adding a test to cover cells with a comma in them. It seems to be that cells with a comma would be surrounded by double quotes.

api/filters.go
Matt-Guest: done

api/filters.go
Matt-Guest: done

preview/preview.go
Matt-Guest: done

PR #45, opened 2017-12-11
Feature/validate filter dimensions

nshumoogum:
### What

Validate filter dimensions and dimension options by calling the dataset API to check whether they exist, if they don't exist then return bad request and message detailing the fields that are wrong. There is future work to return error objects in json instead of a message.

### How to review

Check logic makes sense and that all endpoints (PUT's and POST's) that change a filter blueprint validate request. Any TODO's and FIXME's are to use new endpoints that should be available in future dataset API work.

Make sure unit tests work.

Run acceptance tests by going to `feature/validate-filter-dimensions` on the dp-api-tests branch; then do the following:

1) Run dp-dataset-api
2) Run dp-filter-api (on `feature/validate-filter-dimensions`)
3) cd filterAPI; go test ./...

### Who can review

Team B


models/models.go
```diff
@@ -114,6 +114,69 @@ func (filter *Filter) ValidateFilterBlueprint() error {
...
+type DatasetDimension struct {
+	Name string `bson:"name,omitempty"          json:"dimension,omitempty"`
+}
+
+// ValidateFilterDimensions checks the selected filter dimension
+// are valid for a version of a dataset
+func ValidateFilterDimensions(filterDimensions []Dimension, datasetDimensions *DatasetDimensionResults) error {
+	dimensionNames := make(map[string]string)
+	for _, datasetDimension := range datasetDimensions.Items {
+		dimensionNames[datasetDimension.Name] = datasetDimension.Name
```
gedge: this map (and same in the following func), seems odd - wouldn't `dimensionNames[datasetDimension.Name] = 1` here...

models/models.go
```diff
@@ -114,6 +114,69 @@ func (filter *Filter) ValidateFilterBlueprint() error {
...
+	}
+
+	var incorrectDimensions []string
+	for _, filterDimension := range filterDimensions {
+		if dimensionNames[filterDimension.Name] != filterDimension.Name {
```
gedge: ...and `if _,ok := dimensionNames[filterDimension.Name]; !ok {` here, be a clearer solution?

models/models.go
```diff
@@ -114,6 +114,69 @@ func (filter *Filter) ValidateFilterBlueprint() error {
...
+	Label  string `bson:"label,omitempty"          json:"label"`
+	Option string `bson:"option,omitempty"         json:"option"`
+}
+
+// ValidateFilterDimensionOptions checks the selected filter dimension options
+// are valid for a dimension of a single version of a dataset
+func ValidateFilterDimensionOptions(filterDimensionOptions []string, datasetDimensionOptions *DatasetDimensionOptionResults) []string {
+	dimensionOptions := make(map[string]string)
+	for _, datasetOption := range datasetDimensionOptions.Items {
+		dimensionOptions[datasetOption.Option] = datasetOption.Option
```
gedge: ditto

models/models.go
nshumoogum: Agreed, rushing to get a quick solution out there :/

PR #48, opened 2017-12-21
Feature/healthcheck

gedge:
### What

add healthcheck checking of downstream services (mongo, neo, dataset-api)

### How to review

qv https://github.com/ONSdigital/go-ns/pull/55


cmd/dp-filter-api/main.go
```diff
@@ -99,6 +108,8 @@ func main() {
 
 		cancel()
 
+		healthTicker.Close()
```
CarlHembrough: Consider calling close on the ticker before closing the mongo session.

gedge: moved

dataset/api.go
```diff
@@ -171,6 +172,11 @@ func (api *API) callDatasetAPI(ctx context.Context, method, path string, payload
 	return jsonBody, resp.StatusCode, nil
 }
 
+// GetHealthCheckClient returns a healthcheck-compatible client
+func (api *API) GetHealthCheckClient() *datasetclient.Client {
+	return datasetclient.New(api.url)
```
CarlHembrough: Consider inlining `datasetclient.New(api.url)` in main when the ticker is created.

gedge: As discussed, I was trying to have all `datasetAPI` access in one package, and allow for a possible future merging of this client and the `go-ns` client.  Seemed cleaner to me, so hope we can agree to differ.

PR #52, opened 2018-01-09
Allow events to be passed to the filter api

Matt-Guest:
### What
The filter-api already supported this but there was a bug
which needed to be addressed before events could be added
to a filter output.

Added more information in the swagger spec about completed and
failed jobs.

### How to review
review code and swagger spec

### Who can review
anyone

### DIFF
--- a/mongo/filterstore.go
+++ b/mongo/filterstore.go
@@ -11,12 +11,6 @@ import (
 	"gopkg.in/mgo.v2/bson"
 )
 
-// Filter job states
-const (
-	submitted = "submitted"
-	completed = "completed"
-)
-
 // Error codes
 var (
 	errNotAuthorised = errors.New("Not authorised")
@@ -422,7 +416,7 @@ func createUpdateFilterBlueprint(filter *models.Filter, currentTime time.Time) b
 func createUpdateFilterOutput(filter *models.Filter, currentTime time.Time) bson.M {
 
 	var downloads models.Downloads
-	state := "created"
+	state := models.CreatedState
 	var update bson.M
 	if filter.Downloads != nil {
 		if filter.Downloads.XLS.URL != "" {
@@ -448,7 +442,7 @@ func createUpdateFilterOutput(filter *models.Filter, currentTime time.Time) bson
 			"$set": bson.M{
 				"downloads": downloads,
 				"events":    filter.Events,
-				"state":     completed,
+				"state":     models.CompletedState,
 			},
 			"$setOnInsert": bson.M{
 				"last_updated": currentTime,
@@ -476,7 +470,7 @@ func (s *FilterStore) checkFilterState(filterID string) error {
 		return err
 	}
 
-	if filter.State == submitted {
+	if filter.State == models.SubmittedState {
 		return errForbidden
 	}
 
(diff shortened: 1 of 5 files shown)


mongo/filterstore.go
```diff
@@ -422,19 +422,24 @@ func createUpdateFilterBlueprint(filter *models.Filter, currentTime time.Time) b
 func createUpdateFilterOutput(filter *models.Filter, currentTime time.Time) bson.M {
 
 	var downloads models.Downloads
-
+	state := "created"
 	var update bson.M
+	if filter.Downloads != nil {
+		if filter.Downloads.XLS.URL != "" {
```
daiLlew: is it possible for the Download.XLS to be nil? if so put this inside an if check

mongo/filterstore.go
```diff
@@ -422,19 +422,24 @@ func createUpdateFilterBlueprint(filter *models.Filter, currentTime time.Time) b
 func createUpdateFilterOutput(filter *models.Filter, currentTime time.Time) bson.M {
 
 	var downloads models.Downloads
-
+	state := "created"
```
daiLlew: is there a state constant defined somewhere that we can reuse?

mongo/filterstore.go
```diff
@@ -422,19 +422,24 @@ func createUpdateFilterBlueprint(filter *models.Filter, currentTime time.Time) b
...
 	var update bson.M
+	if filter.Downloads != nil {
+		if filter.Downloads.XLS.URL != "" {
+			downloads.XLS = filter.Downloads.XLS
+		}
 
-	if filter.Downloads.XLS.URL != "" {
-		downloads.XLS = filter.Downloads.XLS
-	}
+		if filter.Downloads.CSV.URL != "" {
```
daiLlew: Same here can Download.CSV be nil?

mongo/filterstore.go
```diff
@@ -422,19 +422,24 @@ func createUpdateFilterBlueprint(filter *models.Filter, currentTime time.Time) b
...
+			downloads.CSV = filter.Downloads.CSV
+		}
 
-	if filter.Downloads.CSV.URL != "" {
-		downloads.CSV = filter.Downloads.CSV
+		if filter.Downloads.JSON.URL != "" {
```
daiLlew: And here

mongo/filterstore.go
nshumoogum: Yes we should probably add states to the models package like we did for dataset API and make them exported constants, then refer to it like so `models.CreatedState`

mongo/filterstore.go
nshumoogum: I think it is good practice to check XLS is not nil.

nshumoogum:
@Matt-Guest - it might be worth adding some unit tests for `mongo.filterstore.go` for the functions that do not call mongo (such as `createUpdateFilterOutput`)

mongo/filterstore.go
Matt-Guest: The Downloads type within a filter-output is a pointer, so it can be nil. Both the CSV and XLSX types within the Downloads are not pointers, so they can not be nil. Checking for nil will trigger a compile error.

mongo/filterstore.go
Matt-Guest: Done 👍

Matt-Guest:
I also added tests around `createUpdateFilterOutput` function.

PR #55, opened 2018-02-07
PUT on /filter-outputs/{id} returns 404 when no auth key is provided

Matt-Guest:
### What

hide authenticated endpoints

### How to review

review the code, run unit tests and acceptance tests

### Who can review

anyone

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,8 @@ build:
 	go build -o $(BUILD_ARCH)/$(BIN_DIR)/dp-filter-api cmd/$(MAIN)/main.go
 debug: build
 	HUMAN_LOG=1 go run -race cmd/$(MAIN)/main.go
+acceptance: build
+	 MONGODB_FILTERS_DATABASE=test HUMAN_LOG=1 go run -race cmd/$(MAIN)/main.go
 test:
 	go test -cover $(shell go list ./... | grep -v /vendor/)
 .PHONY: build debug test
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -13,6 +13,8 @@ build:
 	go build -o $(BUILD_ARCH)/$(BIN_DIR)/dp-filter-api cmd/$(MAIN)/main.go
 debug: build
 	HUMAN_LOG=1 go run -race cmd/$(MAIN)/main.go
+acceptance: build
```
gedge: this is probably a pattern we should use elsewhere (I've been bitten by this, today, on another PR!)

is this documented/used anywhere that you can point to (README should mention it?)?

Makefile
```diff
@@ -13,6 +13,8 @@ build:
...
+	 MONGODB_FILTERS_DATABASE=test HUMAN_LOG=1 go run -race cmd/$(MAIN)/main.go
 test:
 	go test -cover $(shell go list ./... | grep -v /vendor/)
 .PHONY: build debug test
```
gedge: consider adding `acceptance` here

Makefile
gedge: actually, I just noticed that this should probably not depend on `build` as it `run`s it directly

Matt-Guest: I removed it, I also did this for `debug` as its the same command

Makefile
Matt-Guest: Done

Makefile
Matt-Guest: I thrown a section in the readme to cover running the acceptance tests

PR #58, opened 2018-02-20
Add authentication and published flag

eldeal:
### What
- Add middleware to persist authentication in context
- Add 'published' flag to filter blueprints and outputs
- Check for published flag before returning resources

### How to review

Check that unpublished dataset cannot be reached without authentication
Check that filters for published data can still be created as previously
Check that, when authentication is provided, filters can be created for unpublished data

### Who can review

Anyone


eldeal:
I was considering separating the filters.go and api_test.go into separate files, but didn't want to muck up the diffs too much - will probably make that change afterwards (with no other code changes) for clarity but if it would help I could do it sooner.

api/api_test.go
```diff
@@ -114,6 +126,21 @@ func TestFailedToAddFilterBlueprint(t *testing.T) {
...
 
+	Convey("When instance is unpublished and the request is not authenticated, a not found error is returned", t, func() {
+		reader := strings.NewReader(`{"instance_id":"12345678"}`)
+		r, err := http.NewRequest("POST", "http://localhost:22100/filters", reader)
+		So(err, ShouldBeNil)
+
+		w := httptest.NewRecorder()
+		api := routes(authHeader, host, mux.NewRouter(), &mocks.DataStore{}, &mocks.FilterJob{}, &mocks.DatasetAPI{Unpublished: true}, previewMock)
+		api.router.ServeHTTP(w, r)
+		So(w.Code, ShouldEqual, http.StatusBadRequest)
```
nshumoogum: Mismatch in title of test and assertion (e.g. the expected response status), title refers to a not found error but the actual response status in assertion is bad request?

The dataset API will be responding with not found in the near future (PR about to go out)

api/api_test.go
```diff
@@ -347,18 +412,32 @@ func TestFailedToAddFilterBlueprintDimensionOption(t *testing.T) {
...
+		api := routes(authHeader, host, mux.NewRouter(), &mocks.DataStore{Unpublished: true}, &mocks.FilterJob{}, &mocks.DatasetAPI{Unpublished: true}, previewMock)
+		api.router.ServeHTTP(w, r)
+		So(w.Code, ShouldEqual, http.StatusNotFound)
+
+		bodyBytes, _ := ioutil.ReadAll(w.Body)
+		response := string(bodyBytes)
+		So(response, ShouldResemble, "Filter blueprint not found\n")
+	})
+
+	Convey("When a dimension for filter blueprint does not exist, a not found status is returned", t, func() {
```
nshumoogum: The test title does not marry up with assertion, status code is 400 but title assumes it will be 404?

api/api_test.go
```diff
@@ -378,11 +457,25 @@ func TestFailedToAddFilterBlueprintDimensionOption(t *testing.T) {
...
 		So(err, ShouldBeNil)
+
 		w := httptest.NewRecorder()
 		api := routes(authHeader, host, mux.NewRouter(), &mocks.DataStore{}, &mocks.FilterJob{}, &mocks.DatasetAPI{}, previewMock)
+
+		api.router.ServeHTTP(w, r)
+		So(w.Code, ShouldEqual, http.StatusOK)
+	})
+
+	Convey("Successfully get an unpublished filter blueprint with no authentication", t, func() {
```
nshumoogum: Title appears to be wrong, there is authentication on line 473

api/api_test.go
```diff
@@ -1111,4 +1433,14 @@ func TestFailedGetPreview(t *testing.T) {
...
 	})
+
+	Convey("Requesting a preview with no authentication when the instance is unpublished", t, func() {
+		r, err := http.NewRequest("GET", "http://localhost:22100/filter-outputs/21312/preview?limit=a", nil)
+		So(err, ShouldBeNil)
+
+		w := httptest.NewRecorder()
+		api := routes(authHeader, host, mux.NewRouter(), &mocks.DataStore{Unpublished: true}, &mocks.FilterJob{}, &mocks.DatasetAPI{Unpublished: true}, previewMock)
+		api.router.ServeHTTP(w, r)
+		So(w.Code, ShouldEqual, http.StatusBadRequest)
```
nshumoogum: Probably want to check the error message too

mongo/filterstore.go
```diff
@@ -21,9 +21,7 @@ var (
 	errOptionNotFound            = errors.New("Option not found")
 	errFilterOrDimensionNotFound = errors.New("Bad request - filter or dimension not found")
 	errFilterOutputNotFound      = errors.New("Filter output not found")
-
-	errFilterBadRequest    = errors.New("Bad request - filter blueprint not found")
-	errDimensionBadRequest = errors.New("Bad request - filter dimension not found")
+	errFilterBadRequest          = errors.New("Bad request - filter blueprint not found")
```
nshumoogum: I feel like this should not exist, if filter blueprint is not found then it is a 404?

api/api.go
```diff
@@ -99,6 +99,7 @@ type auth struct {
 func (a *auth) authenticate(handle func(http.ResponseWriter, *http.Request)) http.HandlerFunc {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
+
 		if r.Header.Get(internalToken) == a.key {
 			ctx = context.WithValue(ctx, internalToken, true)
```
nshumoogum: golang should not use basic type untyped string as key in context.withvalue

See: https://medium.com/@matryer/context-keys-in-go-5312346a868d

Simply you can do the following:

1) Create ca contextKey variable
```
type contextKey string

func (c contextKey) String() string {
	return string(c)
}
```
2) Replace line 104 with `ctx = context.WithValue(ctx, contextKey(internalToken), true)`

api/filters.go
```diff
@@ -83,16 +100,13 @@ func (api *FilterAPI) addFilterBlueprint(w http.ResponseWriter, r *http.Request)
 	}
 
 	if submitted == "true" {
```
nshumoogum: Ironically this was done by me but if you run `golint ./...`, the result will want the value true to be a constant

Some other things should change such as passing context into functions should be the first parameter (check golint ./...)

models/models.go
```diff
@@ -38,6 +38,7 @@ type Filter struct {
 	Events      Events      `bson:"events,omitempty"     json:"events,omitempty"`
 	FilterID    string      `bson:"filter_id"            json:"filter_id,omitempty"`
 	State       string      `bson:"state,omitempty"      json:"state,omitempty"`
+	Published   bool        `bson:"published,omitempty"   json:"-"`
```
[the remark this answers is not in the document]
CarlHembrough: 👍

PR #67, opened 2018-04-05
Feature/disable private endpoints

CarlHembrough:
### What

* Only enable private endpoints if ENABLE_PRIVATE_ENDPOINTS env var is set
* Default ENABLE_PRIVATE_ENDPOINTS to true in dev env
* Make nomad task names web / publishing specific
* Only enable identity middleware in publishing

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -41,9 +41,16 @@ type FilterAPI struct {
 }
 
 // CreateFilterAPI manages all the routes configured to API
-func CreateFilterAPI(host, bindAddr, zebedeeURL string, datastore DataStore, outputQueue OutputQueue, errorChan chan error, datasetAPI DatasetAPIer, preview PreviewDataset) {
+func CreateFilterAPI(host, bindAddr, zebedeeURL string,
+	datastore DataStore,
+	outputQueue OutputQueue,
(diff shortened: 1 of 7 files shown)


api/api.go
```diff
@@ -79,9 +98,12 @@ func routes(host string, router *mux.Router, dataStore DataStore, outputQueue Ou
 	api.router.HandleFunc("/filters/{filter_blueprint_id}/dimensions/{name}/options/{option}", api.removeFilterBlueprintDimensionOption).Methods("DELETE")
 
 	api.router.HandleFunc("/filter-outputs/{filter_output_id}", api.getFilterOutput).Methods("GET")
-	api.router.HandleFunc("/filter-outputs/{filter_output_id}", api.updateFilterOutput).Methods("PUT")
 	api.router.HandleFunc("/filter-outputs/{filter_output_id}/preview", api.getFilterOutputPreview).Methods("GET")
```
nshumoogum: Is preview just for internal(florence) users and hence should only be available if enablePrivateEndpoints is set to true?

CarlHembrough: I don't think so - I believe this is the preview of the first 20 rows of data, so would be available for both

PR #78, opened 2018-05-11
Return private download links when valid download service token is given

CarlHembrough:
### What

Private links were previously being returned only for authenticated requests. This fails on web now that it does not have authentication. It has been updated to return private links given a valid download service token - the same as the dataset-api

### How to review

Review / test with https://github.com/ONSdigital/dp-api-tests/pull/69

### Who can review

Anyone

### DIFF
--- a/api/filters.go
+++ b/api/filters.go
@@ -39,6 +39,7 @@ var (
 	errDimensionBadRequest   = errors.New("Bad request - filter dimension not found")
 	errDimensionNotFound     = errors.New("Dimension not found")
 	errOptionNotFound        = errors.New("Option not found")
+	errFilterOutputNotFound  = errors.New("Filter output not found")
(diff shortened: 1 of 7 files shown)


api/filters.go
```diff
@@ -940,7 +943,7 @@ func (api *FilterAPI) getOutput(ctx context.Context, filterID string) (*models.F
 	errFilterOutputNotFound := errors.New("Filter output not found")
 
 	// Hide private download links if request is not authenticated
-	if !common.IsCallerPresent(ctx) {
+	if r.Header.Get(common.DownloadServiceHeaderKey) != api.downloadServiceToken {
```
daiLlew: Slightly pedantic but would it be worth logging that we are removing / allowing the links?

CarlHembrough: done 👍 - also improved the existing logging slightly.
 - reuse common logData
 - declare error at package level

PR #79, opened 2018-05-15
Split API handlers into separate go files

CarlHembrough:
### What

All handler functions were previously in filters.go. Now they have been moved into more specifically named files.

 - remove string checking of errors
 - remove capitals / punctuation from errors
 - create common errors to use throughout and prevent cyclic import
 - align error codes with swagger spec (mainly where 400 should be returned instead of 404)

### How to review

Review changes

### Who can review

Anyone


api/filters.go
```diff
@@ -931,63 +337,6 @@ func (api *FilterAPI) getFilter(ctx context.Context, filterID string) (*models.F
...
-			log.Error(err, logData)
-			return nil, errFilterOutputNotFound
-		}
-
-		return output, nil
-	}
-
-	return nil, errFilterOutputNotFound
-}
-
```
eldeal: consider moving the error declarations to the files they're used in. alo the checkFilterOptions function beneath this - should this be in the dimensions/options files?

api/filter_dimensions.go
```diff
@@ -0,0 +1,212 @@
...
+	vars := mux.Vars(r)
+	filterID := vars["filter_blueprint_id"]
+	name := vars["name"]
+	logData := log.Data{
+		"filter_blueprint_id": filterID,
+		"dimension":           name,
+	}
+	log.Info("add filter blueprint dimension", logData)
+
+	options, err := models.CreateDimensionOptions(r.Body)
```
mattrout92: Request body is never closed

api/filters.go
CarlHembrough: The checkFilterOptions function is only used within this file - the name would imply otherwise 😄

PR #89, opened 2018-05-30
Make mongodb updates granular

mattrout92:
### What

- Add a timestamp to avoid race conditions on database updates
- Update unit tests
- Relies on https://github.com/ONSdigital/go-ns/pull/96

### How to review

- Verify logic is correct
- Run unit tests
- Run acceptance tests

### Who can review

Anyone


api/filters.go
```diff
@@ -349,6 +354,8 @@ func (api *FilterAPI) updateFilterBlueprint(ctx context.Context, filter *models.
 		}
 	}
 
+	newFilter.UniqueTimestamp = timestamp
```
nshumoogum: Again i don't think we should be setting the unique timestamp for new filter with the old timestamp (I know this gets reset on insert to datastore but is confusing to read).

The previous timestamp should be passed into UpdateFilter method on datastore package and used in the select method.

If the WithUpdates function in go-ns changed there is a possibility for this to break or if not at very least mislead investigation into our mongo documents when we have real issues.

mongo/filterstore.go
```diff
@@ -40,6 +41,9 @@ func (s *FilterStore) AddFilter(host string, filter *models.Filter) (*models.Fil
 	session := s.Session.Copy()
 	defer session.Close()
 
+	// Initialise with a timestamp
+	filter.UniqueTimestamp = 1
```
nshumoogum: You can create a valid mongo timestamp with gedges mgo bson package by doing the following:

`filter.UniqueTimestamp = bson.NewMongoTimestamp(currentTime, 1)`

mongo/filterstore.go
```diff
@@ -209,11 +228,7 @@ func (s *FilterStore) CreateFilterOutput(filter *models.Filter) error {
 	session := s.Session.Copy()
 	defer session.Close()
 
-	if err := session.DB(s.db).C(s.outputsCollection).Insert(filter); err != nil {
-		return err
-	}
-
-	return nil
+	return session.DB(s.db).C(s.outputsCollection).Insert(filter)
```
nshumoogum: We need to create a valid mongo timestamp on insert, this can be done by the following prior to return statement

`filter.UniqueTimestamp = bson.NewMongoTimestamp(currentTime, 1)`

nshumoogum:
So as you are ready to crack on with updating this, could you strip out the checks on state for filter blueprint endpoints/methods as they no longer exist.

I assume the unit tests will break as there will checks for invalid status', should be as simple as removing those unique unit tests.

mongo/filterstore.go
```diff
@@ -72,12 +76,15 @@ func (s *FilterStore) UpdateFilter(updatedFilter *models.Filter) error {
 	session := s.Session.Copy()
 	defer session.Close()
 
-	update := createUpdateFilterBlueprint(updatedFilter, time.Now())
+	update, err := mongolib.WithUpdates(createUpdateFilterBlueprint(updatedFilter, time.Now()))
```
nshumoogum: Should not be passing in time.Now() as all timestamps are handled by `mongolib.WithUpdates`.

This means removing all instances within mongo filestore.go where the `last_updated` field gets setOnInsert as we should always be relying on the `mongolib.WithUpdates` function to handle this for us.

mongo/filterstore.go
```diff
@@ -126,12 +133,15 @@ func (s *FilterStore) AddFilterDimension(filterID, name string, options []string
 		list = append(list, d)
 	}
 
-	queryFilter := bson.M{"filter_id": filterID}
-	update := bson.M{"$set": bson.M{"dimensions": list}}
+	queryFilter := bson.M{"filter_id": filterID, "unique_timestamp": timestamp}
```
nshumoogum: `queryFilter` would be better off being called `selectFilter` or even better to be just `select`; the Update query includes select and update parts

This can be applied to all update queries in filterstore

PR #90, opened 2018-05-30
Feature/audit filter options

CarlHembrough:
### What

 - refactor business logic out of handler function, reducing the surface of auditing locations
 - add auditing to filter dimension endpoints
 - refactor tests to use nested conveys, saving common setup code and grouping common tests
 - rework some error handling to support existing response codes / messages

### How to review

Review / test

### Who can review

Anyone


api/filter_dimension_options.go
```diff
@@ -54,10 +56,19 @@ func (api *FilterAPI) getFilterBlueprintDimensionOptions(w http.ResponseWriter,
 	bytes, err := json.Marshal(options)
 	if err != nil {
 		log.ErrorC("failed to marshal filter blueprint dimension options into bytes", err, logData)
+		if auditErr := api.auditor.Record(r.Context(), getOptionsAction, actionUnsuccessful, auditParams); auditErr != nil {
```
daiLlew: This is just an opinion BUT - if you changed the return of your function to `([]byte, error)` then you could return the marshalled bytes and save yourself this extra audit unsuccessful step.

api/filter_dimension_options.go
```diff
@@ -108,141 +179,174 @@ func (api *FilterAPI) getFilterBlueprintDimensionOption(w http.ResponseWriter, r
...
 
-	if filterBlueprint.State == models.SubmittedState {
-		log.Error(errForbidden, logData)
-		setErrorCode(w, errForbidden, filterBlueprint.State)
+	if auditErr := api.auditor.Record(r.Context(), addOptionAction, actionSuccessful, auditParams); auditErr != nil {
+		logAuditFailure(r.Context(), addOptionAction, actionSuccessful, auditErr, logData)
 		return
 	}
 
+	setJSONContentType(w)
```
daiLlew: If audit successful  fails do you want to let it fall through to allow it to set the status and return type? In this scenario the dimension has already added so there is no harm in returning the appropriate success response.

api/filter_dimension_options.go
```diff
@@ -108,141 +179,174 @@ func (api *FilterAPI) getFilterBlueprintDimensionOption(w http.ResponseWriter, r
...
 
-	filterBlueprint, err := api.getFilterBlueprint(r.Context(), filterID)
+	auditParams := common.Params{
+		"filter_blueprint_id": filterBlueprintID,
+		"dimension":           dimensionName,
+		"option":              option,
+	}
+	if auditErr := api.auditor.Record(r.Context(), removeOptionAction, actionAttempted, auditParams); auditErr != nil {
+		handleAuditingFailure(r.Context(), removeOptionAction, actionAttempted, w, auditErr, logData)
+		return
```
daiLlew: Should we be setting an error status code in this case?

api/filter_dimension_options.go
CarlHembrough: Yes you're right - I shouldn't have the return after the `logAuditFailure` 👍

api/filter_dimension_options.go
CarlHembrough: `handleAuditingFailure` will set a 500 error if the audit fails. I have removed the return statement in this function below `logAuditFailure` (I wonder if this is what you referring to?)

PR #91, opened 2018-05-31
Remove duplicate options that are sent by client

mattrout92:
### What

There is currently a bug where if you post a list of dimension options to the filter api, it will accept duplicate options: https://api.beta.ons.gov.uk/v1/filters/8af54449-7200-42ed-a778-a789e1717ab5/dimensions

This is a fix for that issue

### How to review

- Verify logic is correct

### Who can review

Anyone


api/filter_dimensions.go
```diff
@@ -221,3 +223,17 @@ func (api *FilterAPI) checkNewFilterDimension(ctx context.Context, name string,
 
 	return nil
 }
+
+func removeDuplicates(elements []string) []string {
```
nshumoogum: Maybe name could be more specific? removeDuplicateOptions

Worth adding a couple of unit tests for this function

nshumoogum:
Also you could say there is another bug on creating a filter blueprint as it allows duplicate dimensions; which could obtain different options; like so:

```
...
"dimensions": [
    {
      "name": "age",
      "options": [
        "24",
        "26"
      ]
    },
    {
      "name": "age",
      "options": [
        "25",
        "33"
      ]
    }
  ],
...
```

api/filter_dimensions.go
mattrout92: This is deliberately generic so that it can be reused for other things like dimensions.

models/models.go
```diff
@@ -114,6 +115,15 @@ var (
 	ErrorNoData      = errors.New("Bad request - Missing data in body")
 )
 
+// ErrorDuplicateDimension is returned if a request contains a duplicate dimension
+type ErrorDuplicateDimension struct {
```
nshumoogum: Naming convention of error types should end with `Error` - see [here](https://github.com/golang/go/wiki/Errors#naming)

PR #92, opened 2018-06-01
Refactor to use go-ns dataset API client

CarlHembrough:
### What

Refactor to use go-ns dataset API client

### How to review

Review / test 

### Who can review

Anyone


api/api.go
```diff
@@ -14,19 +15,22 @@ import (
 	"github.com/justinas/alice"
 )
 
-var httpServer *server.Server
+//go:generate moq -out datastoretest/preview.go -pkg datastoretest . PreviewDataset
 
-type key string
+var httpServer *server.Server
```
mattrout92: It feels strange that this is a global variable?

PR #98, opened 2018-06-12
Feature/output event endpoint

CarlHembrough:
### What

Add new endpoints allowing events to be added to a filter output

### How to review

Review code / swagger spec / test with https://github.com/ONSdigital/dp-api-tests/pull/85

### Who can review

Anyone

### DIFF
--- a/api/filter_outputs.go
+++ b/api/filter_outputs.go
@@ -16,6 +16,7 @@ import (
 	"github.com/ONSdigital/dp-filter-api/preview"
 	"github.com/ONSdigital/go-ns/common"
 	"github.com/pkg/errors"
+	"io/ioutil"
 	"time"
 )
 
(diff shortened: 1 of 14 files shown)


api/filter_outputs.go
```diff
@@ -418,3 +419,50 @@ func buildDownloadsObject(previousFilterOutput, filterOutput *models.Filter, dow
 		}
 	}
 }
+
+func (api *FilterAPI) addEventHandler(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
nshumoogum: I think vendor needs to be updated for all dependencies from go-ns otherwise the above won't be possible

api/filter_outputs.go
```diff
@@ -418,3 +419,50 @@ func buildDownloadsObject(previousFilterOutput, filterOutput *models.Filter, dow
...
+	vars := mux.Vars(r)
+	filterOutputID := vars["filter_output_id"]
+
+	logData := log.Data{"filter_output_id": filterOutputID}
+	log.Info("add event to filter output endpoint called", logData)
+
+	defer r.Body.Close()
+
+	bytes, err := ioutil.ReadAll(r.Body)
+	if err != nil {
```
nshumoogum: Do we not audit action?

api/filter_outputs.go
```diff
@@ -418,3 +419,50 @@ func buildDownloadsObject(previousFilterOutput, filterOutput *models.Filter, dow
...
+
+	err = api.addEvent(filterOutputID, event)
+	if err != nil {
+		log.Error(errors.Wrap(err, "failed to add event to filter output"), logData)
+		setErrorCode(w, err)
+		return
+	}
+
+	w.WriteHeader(http.StatusCreated)
+	log.Debug("added event to filter output", logData)
```
nshumoogum: Think about using Info instead of Debug

api/filter_outputs.go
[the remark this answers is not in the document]
CarlHembrough: I have updated go-ns dependencies and replaced `log.Error()` with `log.ErrorCtx()`, and `log.Info()` with `log.InfoCtx()`

api/filter_outputs.go
CarlHembrough: I considered adding auditing but realised this endpoint will never be called with a user. It is only called from the exporter services which are kafka driven and have no auth. Do you think its worth adding for completeness?

api/filter_outputs.go
CarlHembrough: Done (using `InfoCtx` 😉 )

api/filter_outputs.go
nshumoogum: So I thought as much and was thinking maybe its worth adding just so its all tied up neatly and probably more importantly is that currently this is still accessible to a florence user until we add permissions to users.

Lastly we do not know what services (maybe api's) might call this endpoint in the future.

Maybe open a discussion with ELeanor and Ian (when he gets back from holiday).

I will approve for now though.

PR #105, opened 2018-07-16
Feature/drain request bodies

CarlHembrough:
### What

* Add tests to ensure request bodies are drained
* Update go-ns identity packages and required dependencies, so that request bodies are drained in the identity middleware / check wrapper function

### How to review

Review changes / test

### Who can review

Anyone


api/filter_dimensions_test.go
```diff
@@ -203,6 +204,11 @@ func TestSuccessfulAddFilterBlueprintDimension(t *testing.T) {
 		Convey("Then the auditor is called for the attempt and outcome", func() {
 			assertAuditCalled(mockAuditor, addDimensionAction, actionSuccessful, expectedAuditParams)
 		})
+
+		Convey("Then the request body has been drained", func() {
+			_, err = r.Body.Read(make([]byte, 1))
```
nshumoogum: Possibly check body is nil instead of relying on there being an error, if there is no error you may want to know whether the body is nil or not as this will give you more to work with if it breaks.

CarlHembrough: I have updated the test to also check the first returned value - the number of bytes read.

PR #127, opened 2019-02-28
Update preview to use graph abstraction and update relevant tests

eldeal:
### What

Update preview to use graph abstraction and update relevant tests

### How to review

check preview returns results

### Who can review

anyone


cmd/dp-filter-api/main.go
```diff
@@ -91,13 +85,12 @@ func main() {
...
 	datasetAPI := dataset.NewAPIClient(cfg.DatasetAPIURL, cfg.ServiceAuthToken, "")
 
-	observationStore := observation.NewStore(pool)
 	previewDatasets := preview.DatasetStore{Store: observationStore}
 	outputQueue := filterOutputQueue.CreateOutputQueue(producer.Output())
 
 	healthTicker := healthcheck.NewTicker(
 		cfg.HealthCheckInterval,
-		neo4jhealth.NewHealthCheckClient(pool),
+		//	observationStore.Healthcheck(),
```
jondewijones: Do we want this commented out?

cmd/dp-filter-api/main.go
```diff
@@ -54,15 +52,11 @@ func main() {
 		os.Exit(1)
 	}
 
-	// Driver pool will never return an error as a bolt connection is never created. So we test it by creating
-	// a connection.
-	pool, _ := bolt.NewClosableDriverPool(cfg.Neo4jURL, cfg.Neo4jPoolSize)
-	conn, err := pool.OpenPool()
+	observationStore, err := graph.New(context.Background(), graph.Subsets{Observation: true})
 	if err != nil {
 		log.ErrorC("could not connect to neo4j", err, nil)
```
jondewijones: Should this be a bit more generic for when we're not using Neo?

cmd/dp-filter-api/main.go
eldeal: will be in a coming PR, in order to healthcheck this in abstraction-land we'll need to completely change the way healthchecks work so felt a significantly difference change that's worth separating

PR #148, opened 2020-04-03
Feature/healthcheck

CarlHembrough:
### What
- Migrate solution to use go modules
- Upgrade to go version 1.13.8
- Use dp-healthcheck/healthcheck to healthcheck
    - Kafka
    - MongoDB
    - Zebedee
    - Dataset API
    - Neo4j
(The download service was originally listed on the task to be healthchecked. See the note added to the task: https://trello.com/c/smqVA3rG/4623-replace-implement-health-check-in-dp-filter-api-that-complies-with-new-spec-35d)

- Migrate from vendor to go modules
- Migrate to use log.go instead of go-ns/log
- Migrate to use ONSdigital/dp-kafka
- Migrate to use ONSdigital/dp-api-clients-go/identity

### How to review
- Review code
- Unit test
- Test health endpoint curl -v localhost:22100/health
- Integration test with other services

### Who can review
Anyone


go.mod
```diff
@@ -0,0 +1,24 @@
+module github.com/ONSdigital/dp-filter-api
+
+go 1.14
+
+require (
+	github.com/ONSdigital/dp-api-clients-go v1.9.0
+	github.com/ONSdigital/dp-graph v1.0.2
```
DavidSubiros, line 7: I think it would be worth moving to dp-graph version v2.0.3, although this might require some changes in the models?
If we update this dependency, we will need to use the v2 path (required by go modules). I.e.:

in go.mod
```
require(
    ...
    github.com/ONSdigital/dp-graph/v2 v2.0.3
    ...
```

imports in code:
```
import (
    ...
    "github.com/ONSdigital/dp-graph/v2/...
```

Example of similar usage in: dp-dimension-importer, dp-code-list-api and dp-hierarchy-builder.

config/config.go
```diff
@@ -52,6 +53,7 @@ func Get() (*Config, error) {
 		DatasetAPIURL:              "http://localhost:22000",
 		DatasetAPIAuthToken:        "FD0108EA-825D-411C-9B1D-41EF7727F465",
 		HealthCheckInterval:        30 * time.Second,
```
DavidSubiros, line 55: Can we set the default values to 30s and 90s, please?
```
		HealthCheckInterval:        30 * time.Second,
		HealthCheckCriticalTimeout: 90 * time.Second,
```

cmd/dp-filter-api/main.go
```diff
@@ -3,94 +3,106 @@ package main
...
+	logIfError(ctx, err, "error creating kafka filter output submitted producer")
+	producer.Channels().LogErrors(ctx, "error received from kafka producer, topic: "+cfg.FilterOutputSubmittedTopic)
 
 	var auditor audit.AuditorService
-	var auditProducer kafka.Producer
+	var auditProducer *kafka.Producer
 
 	if cfg.EnablePrivateEndpoints {
-		log.Info("private endpoints enabled, enabling auditing", nil)
+		log.Event(ctx, "private endpoints enabled, enabling auditing")
```
DavidSubiros, line 76: can we add severity log.INFO?

cmd/dp-filter-api/main.go
```diff
@@ -3,94 +3,106 @@ package main
...
-		logIfError(err, "error creating kafka audit producer")
+		logIfError(ctx, err, "error creating kafka audit producer")
+		auditProducer.Channels().LogErrors(ctx, "error received from kafka producer, topic: "+cfg.AuditEventsTopic)
 
-		auditor = audit.New(auditProducer, "dp-filter-api")
+		auditProducerAdapter := kafkaadapter.NewProducerAdapter(auditProducer)
+		auditor = audit.New(auditProducerAdapter, "dp-filter-api")
 	} else {
-		log.Info("private endpoints disabled, auditing will not be enabled", nil)
+		log.Event(ctx, "private endpoints disabled, auditing will not be enabled")
```
DavidSubiros, line 91: can we add severity log.INFO?

cmd/dp-filter-api/main.go
```diff
@@ -101,79 +113,125 @@ func main() {
 		cfg.EnablePrivateEndpoints,
 		cfg.DownloadServiceURL,
 		cfg.DownloadServiceSecretKey,
+		cfg.ServiceAuthToken,
 		auditor,
+		&hc,
 	)
```
DavidSubiros, line 120: even if we no longer kill the service on kafka errors, we could log them like so:
```
producer.Channels().LogErrors(ctx, "kafka producer")
auditProducer.Channels().LogErrors(ctx, "kafka audit producer")
```

cmd/dp-filter-api/main.go
```diff
@@ -101,79 +113,125 @@ func main() {
...
 		if serviceList.AuditProducer {
-			log.Info("closing audit producer", nil)
-			logIfError(auditProducer.Close(ctx), "unable to close audit producer")
+			log.Event(ctx, "closing audit producer")
+			logIfError(ctx, auditProducer.Close(ctx), "unable to close audit producer")
 		}
 	}()
 
 	// wait for shutdown success (via cancel) or failure (timeout)
 	<-ctx.Done()
```
DavidSubiros, line 166: We could log when a shutdown times out:
```
	if ctx.Err() == context.DeadlineExceeded {
		log.Event(ctx, "shutdown timeout", log.ERROR, log.Error(ctx.Err()))
		os.Exit(1)
	}
```

cmd/dp-filter-api/main.go
```diff
@@ -101,79 +113,125 @@ func main() {
...
 
-	log.Info("Shutdown complete", nil)
+	log.Event(ctx, "Shutdown complete")
 	os.Exit(1)
```
DavidSubiros, line 168: if we existed with err 1 on timeout, here we can do `os.Exit(0)`

cmd/dp-filter-api/main.go
```diff
@@ -101,79 +113,125 @@ func main() {
...
 }
 
-func exitIfError(err error, message string) {
+func startHealthCheck(ctx context.Context, cfg *config.Config, datasetAPI *dataset.Client, producer *kafka.Producer, observationStore *graph.DB, dataStore *mongo.FilterStore, auditProducer *kafka.Producer) healthcheck.HealthCheck {
```
DavidSubiros, line 171: I like this approach of having the creation and register checkers in a function outside main func :)

cmd/dp-filter-api/main.go
```diff
@@ -101,79 +113,125 @@ func main() {
...
+	}
+
+	hc.Start(ctx)
+	return hc
+}
+
+func exitIfError(ctx context.Context, err error, message string) {
 	if err != nil {
-		log.ErrorC(message, err, nil)
+		log.Event(ctx, message, log.Error(err))
```
DavidSubiros, line 228: can we add severity log.FATAL?

cmd/dp-filter-api/main.go
```diff
@@ -101,79 +113,125 @@ func main() {
...
 		os.Exit(1)
 	}
 }
 
-func logIfError(err error, message string) {
+func logIfError(ctx context.Context, err error, message string) {
 	if err != nil {
-		log.ErrorC(message, err, nil)
+		log.Event(ctx, message, log.Error(err))
```
DavidSubiros, line 235: can we add severity log.ERROR?

api/filter_dimensions.go
```diff
@@ -236,22 +237,23 @@ func (api *FilterAPI) addFilterBlueprintDimensionHandler(w http.ResponseWriter,
...
+		handleAuditingFailure(ctx, addDimensionAction, actionAttempted, w, auditErr, logData)
 		return
 	}
 
 	options, err := models.CreateDimensionOptions(r.Body)
 	if err != nil {
-		log.ErrorC("unable to unmarshal request body", err, logData)
-		if auditErr := api.auditor.Record(r.Context(), addDimensionAction, actionUnsuccessful, auditParams); auditErr != nil {
-			handleAuditingFailure(r.Context(), addDimensionAction, actionUnsuccessful, w, auditErr, logData)
+		log.Event(ctx, "unable to unmarshal request body", log.Error(err), logData)
```
DavidSubiros, line 254: can we add severity log.ERROR?

api/filter_dimensions.go
```diff
@@ -338,14 +340,14 @@ func (api *FilterAPI) addFilterBlueprintDimension(ctx context.Context, filterBlu
 func (api *FilterAPI) checkNewFilterDimension(ctx context.Context, name string, options []string, dataset *models.Dataset) error {
 
 	logData := log.Data{"dimension_name": name, "dimension_options": options, "dataset": dataset}
-	log.InfoCtx(ctx, "check filter dimensions and dimension options before calling api, see version number", logData)
+	log.Event(ctx, "check filter dimensions and dimension options before calling api, see version number", logData)
```
DavidSubiros, line 343: can we add severity log.INFO?

api/filter_outputs.go
```diff
@@ -138,25 +139,25 @@ func (api *FilterAPI) updateFilterOutputHandler(w http.ResponseWriter, r *http.R
 func (api *FilterAPI) updateFilterOutput(ctx context.Context, filterOutputID string, filterOutput *models.Filter) error {
 
 	logData := log.Data{"filter_output_id": filterOutputID}
-	log.InfoCtx(ctx, "updating filter output", logData)
+	log.Event(ctx, "updating filter output", logData)
```
DavidSubiros, line 142: can we add severity log.INFO?

api/filter_outputs.go
```diff
@@ -174,27 +175,27 @@ func (api *FilterAPI) updateFilterOutput(ctx context.Context, filterOutputID str
 
 	isNowStatusCompleted := false
 	if downloadsAreGenerated(filterOutput) {
-		log.InfoCtx(ctx, "downloads have been generated, setting filter output status to completed", logData)
+		log.Event(ctx, "downloads have been generated, setting filter output status to completed", logData)
```
DavidSubiros, line 178: can we add severity log.INFO?

api/filter_outputs.go
```diff
@@ -174,27 +175,27 @@ func (api *FilterAPI) updateFilterOutput(ctx context.Context, filterOutputID str
...
 	if err = api.dataStore.UpdateFilterOutput(filterOutput, timestamp); err != nil {
-		log.ErrorCtx(ctx, errors.Wrap(err, "unable to update filter output"), logData)
+		log.Event(ctx, "unable to update filter output", log.ERROR, log.Error(err), logData)
 		return err
 	}
 
 	// save the completed event after saving the filter output if its now complete
 	if isNowStatusCompleted {
-		log.InfoCtx(ctx, "filter output status is now completed, creating completed event", logData)
+		log.Event(ctx, "filter output status is now completed, creating completed event", logData)
```
DavidSubiros, line 190: can we add severity log.INFO?

api/filter_outputs.go
```diff
@@ -263,57 +265,57 @@ func (api *FilterAPI) getFilterOutputPreviewHandler(w http.ResponseWriter, r *ht
...
 
-func (api *FilterAPI) getFilterOutputPreview(ctx context.Context, filterOutputID string, limit int) (*preview.FilterPreview, error) {
+func (api *FilterAPI) getFilterOutputPreview(ctx context.Context, filterOutputID string, limit int) (*models.FilterPreview, error) {
 
 	logData := log.Data{
 		"filter_output_id": filterOutputID,
 		"limit":            limit,
 	}
-	log.InfoCtx(ctx, "get filter output preview", logData)
+	log.Event(ctx, "get filter output preview", logData)
```
DavidSubiros, line 300: can we add severity log.INFO?

api/filter_outputs.go
```diff
@@ -326,7 +328,7 @@ func (api *FilterAPI) getOutput(ctx context.Context, filterID string, hideS3Link
 
 	output, err := api.dataStore.GetFilterOutput(filterID)
 	if err != nil {
-		log.ErrorCtx(ctx, err, logData)
+		log.Event(ctx, "error getting filter output", log.Error(err), logData)
```
DavidSubiros, line 331: can we add severity log.ERROR?

api/filter_outputs.go
```diff
@@ -335,7 +337,7 @@ func (api *FilterAPI) getOutput(ctx context.Context, filterID string, hideS3Link
 	// Hide private download links if request is not authenticated
 	if hideS3Links {
 
-		log.InfoCtx(ctx, "a valid download service token has not been provided. hiding links", logData)
+		log.Event(ctx, "a valid download service token has not been provided. hiding links", logData)
```
DavidSubiros, line 340: can we add severity log.INFO?

api/filter_outputs.go
```diff
@@ -348,27 +350,27 @@ func (api *FilterAPI) getOutput(ctx context.Context, filterID string, hideS3Link
 			}
 		}
 	} else {
-		log.InfoCtx(ctx, "a valid download service token has been provided. not hiding private links", logData)
+		log.Event(ctx, "a valid download service token has been provided. not hiding private links", logData)
```
DavidSubiros, line 353: can we add severity log.INFO?

api/filter_outputs.go
```diff
@@ -348,27 +350,27 @@ func (api *FilterAPI) getOutput(ctx context.Context, filterID string, hideS3Link
...
 	}
 
 	//only return the filter if it is for published data or via authenticated request
 	if output.Published != nil && *output.Published == models.Published || common.IsCallerPresent(ctx) {
 		return output, nil
 	}
 
-	log.InfoCtx(ctx, "unauthenticated request to access unpublished filter output", logData)
+	log.Event(ctx, "unauthenticated request to access unpublished filter output", logData)
```
DavidSubiros, line 361: can we add severity log.INFO?

api/filter_outputs.go
```diff
@@ -348,27 +350,27 @@ func (api *FilterAPI) getOutput(ctx context.Context, filterID string, hideS3Link
...
+		log.Event(ctx, "failed to retrieve filter blueprint", log.ERROR, log.Error(err), logData)
 		return nil, filters.ErrFilterOutputNotFound
 	}
 
 	//filter has been published since output was last requested, so update output and return
 	if filter.Published != nil && *filter.Published == models.Published {
 		output.Published = &models.Published
 		if err := api.dataStore.UpdateFilterOutput(output, output.UniqueTimestamp); err != nil {
-			log.ErrorCtx(ctx, err, logData)
+			log.Event(ctx, "error updating filter output", log.Error(err), logData)
```
DavidSubiros, line 373: can we add severity log.ERROR?

cmd/dp-filter-api/main.go
```diff
@@ -101,79 +113,125 @@ func main() {
...
-		case err := <-apiErrors:
-			log.ErrorC("api error received", err, nil)
-		}
-	}()
-
 	// block until a fatal error occurs
 	select {
 	case <-signals:
-		log.Info("os signal received", nil)
+		log.Event(ctx, "os signal received")
```
DavidSubiros, line 124: can we add severity log.INFO?

cmd/dp-filter-api/main.go
```diff
@@ -101,79 +113,125 @@ func main() {
...
-			logIfError(err, "unable to close api server")
+			logIfError(ctx, err, "unable to close api server")
 		}
 
-		healthTicker.Close()
+		hc.Stop()
 
 		if serviceList.FilterStore {
-			log.Info("closing filter store", nil)
+			log.Event(ctx, "closing filter store")
```
DavidSubiros, line 141: can we add severity log.INFO?

cmd/dp-filter-api/main.go
```diff
@@ -101,79 +113,125 @@ func main() {
...
 			// mongo.Close() may use all remaining time in the context
-			logIfError(mongolib.Close(ctx, dataStore.Session), "unable to close filter store")
+			logIfError(ctx, mongolib.Close(ctx, dataStore.Session), "unable to close filter store")
 		}
 
 		if serviceList.ObservationStore {
-			log.Info("closing observation store", nil)
-			logIfError(observationStore.Close(ctx), "unable to close observation store")
+			log.Event(ctx, "closing observation store")
```
DavidSubiros, line 147: can we add severity log.INFO?

cmd/dp-filter-api/main.go
```diff
@@ -101,79 +113,125 @@ func main() {
...
+			logIfError(ctx, observationStore.Close(ctx), "unable to close observation store")
 		}
 
 		if serviceList.FilterOutputSubmittedProducer {
-			log.Info("closing filter output submitted producer", nil)
+			log.Event(ctx, "closing filter output submitted producer")
```
DavidSubiros, line 152: can we add severity log.INFO?

cmd/dp-filter-api/main.go
```diff
@@ -101,79 +113,125 @@ func main() {
...
 			// Close producer after http server has closed so if a message
 			// needs to be sent to kafka off a request it can
-			logIfError(producer.Close(ctx), "unable to close filter output submitted producer")
+			logIfError(ctx, producer.Close(ctx), "unable to close filter output submitted producer")
 		}
 
 		if serviceList.AuditProducer {
-			log.Info("closing audit producer", nil)
-			logIfError(auditProducer.Close(ctx), "unable to close audit producer")
+			log.Event(ctx, "closing audit producer")
```
DavidSubiros, line 159: can we add severity log.INFO?

cmd/dp-filter-api/main.go
```diff
@@ -101,79 +113,125 @@ func main() {
...
+			logIfError(ctx, auditProducer.Close(ctx), "unable to close audit producer")
 		}
 	}()
 
 	// wait for shutdown success (via cancel) or failure (timeout)
 	<-ctx.Done()
 
-	log.Info("Shutdown complete", nil)
+	log.Event(ctx, "Shutdown complete")
```
DavidSubiros, line 167: can we add severity log.INFO?

cmd/dp-filter-api/main.go
```diff
@@ -101,79 +113,125 @@ func main() {
...
+	versionInfo, err := healthcheck.NewVersionInfo(BuildTime, GitCommit, Version)
+	if err != nil {
+		log.Event(ctx, "error creating version info", log.FATAL, log.Error(err))
+		hasErrors = true
+	}
+
+	hc := healthcheck.New(versionInfo, cfg.HealthCheckCriticalTimeout, cfg.HealthCheckInterval)
+
+	if err = hc.AddCheck("Dataset API", datasetAPI.Checker); err != nil {
+		log.Event(ctx, "error creating dataset API health check", log.Error(err))
```
DavidSubiros, line 184: can we add severity log.ERROR?

api/filter_dimensions.go
```diff
@@ -27,19 +26,20 @@ func (api *FilterAPI) getFilterBlueprintDimensionsHandler(w http.ResponseWriter,
 	vars := mux.Vars(r)
 	filterBlueprintID := vars["filter_blueprint_id"]
 	logData := log.Data{"filter_blueprint_id": filterBlueprintID}
```
nshumoogum, line 28: So in the dataset API I added an `action` variable to logData as I believe it is useful when debugging environments and understanding what went wrong with that service. Would it be worth doing the same in the filter API?

For example here:
`logData := log.Data{"filter_blueprint_id": filterBlueprintID, "action": getDimensionsAction}`

api/filters.go
```diff
@@ -166,21 +165,21 @@ func (api *FilterAPI) createFilterBlueprint(ctx context.Context, filter *models.
...
+	if _, err = api.dataStore.AddFilter(newFilter); err != nil {
+		log.Event(ctx, "failed to create new filter blueprint", log.ERROR, log.Error(err), logData)
 		return nil, err
 	}
 
 	if submitted == filterSubmitted {
 		var filterOutput models.Filter
 		// Create filter output resource and use id to pass into kafka
-		filterOutput, err = api.createFilterOutputResource(newFilter, newFilter.FilterID)
+		filterOutput, err = api.createFilterOutputResource(nil, newFilter, newFilter.FilterID)
```
nshumoogum, line 180: Missing context value

api/filters.go
```diff
@@ -492,7 +493,15 @@ func (api *FilterAPI) checkFilterOptions(ctx context.Context, newFilter *models.
 
 func (api *FilterAPI) getVersion(ctx context.Context, dataset *models.Dataset) (*datasetAPI.Version, error) {
 
-	dimensions, err := api.datasetAPI.GetVersion(ctx, dataset.ID, dataset.Edition, strconv.Itoa(dataset.Version))
+	dimensions, err := api.datasetAPI.GetVersion(ctx,
+		"",
```
nshumoogum, line 497: Should we not be passing user auth token in here? If the request is from a user then we should be using the user Auth token to validate if user can retrieve this information?

I think getVersion function signature should contain the user auth token and possibly collection id too so that it can be used when calling `api.datasetAPI.GetVersion(...` func if it exists in `r.Context()`

nshumoogum: Ah just seen you do exactly this on line 543 and 545, so why not here?

ci/scripts/build.sh
```diff
@@ -2,9 +2,7 @@
 
 cwd=$(pwd)
 
-export GOPATH=$cwd/go
-
-pushd $GOPATH/src/github.com/ONSdigital/dp-filter-api
+pushd $cwd/dp-filter-api
   make build && mv build/$(go env GOOS)-$(go env GOARCH)/* $cwd/build
```
nshumoogum, line 6: I dont think you need the GOOS and GOARCH environment variables anymore, see [dp-dataset API pr](https://github.com/ONSdigital/dp-dataset-api/pull/215)

cmd/dp-filter-api/main.go
nshumoogum, line 124: I think we can add the apierrors case back in here, if there is an issue with the server force a graceful shutdown as the application will be unable to process anything

cmd/dp-filter-api/main.go
```diff
@@ -101,79 +113,125 @@ func main() {
...
+		// zebedee is used only for identity checking
+		zebedeeClient := zebedee.New(cfg.ZebedeeURL)
+		if err = hc.AddCheck("Zebedee", zebedeeClient.Checker); err != nil {
+			log.Event(ctx, "error creating zebedee health check", log.ERROR, log.Error(err))
+			hasErrors = true
+		}
+	}
+
+	if hasErrors {
+		os.Exit(1)
```
nshumoogum, line 219: maybe return error and handle this in call to this function

cmd/dp-filter-api/main.go
nshumoogum, line 228: Missing severity level, `log.FATAL`

config/config.go
```diff
@@ -52,6 +53,7 @@ func Get() (*Config, error) {
...
+		HealthCheckCriticalTimeout: 2 * time.Second,
```
nshumoogum, line 56: If we the default value to 90 seconds please

go.mod
```diff
@@ -0,0 +1,24 @@
+module github.com/ONSdigital/dp-filter-api
+
+go 1.14
```
nshumoogum, line 3: This should be on go 1.13

cmd/dp-filter-api/main.go
CarlHembrough: I have added these lines on line 70 / 86 of main.go - just after the initialisation of the producers.

cmd/dp-filter-api/main.go
CarlHembrough: I agree - though in other services we have been just logging when a server error occurs, and not shutting down the service?

cmd/dp-filter-api/main.go
CarlHembrough: I considered that, though I thought the code would be cleaner like this, as we would have to create a new error instance to return and we can maintain just one return value?

api/filters.go
CarlHembrough: This was a tricky part of the changes - It wasn't clear when I needed it and when I didn't. I have aimed to replicate exactly what was happening before. I will double check that the changes reflect the previous behaviour.

cmd/dp-filter-api/main.go
nshumoogum: Yes some because those applications have multiple responsibilities for instance import pipeline apps are predominantly event driven via kafka not by http requests, they are only have a http server for healthcheck endpoint.

If those apps http server fails then they maybe able to continue processing data and the platform can decide what to do (e.g. kill the service if it doesn't like the 500 response it will receive if the server is not working)

nshumoogum: In this case the API is reliant on having a server and therefore any issues should force the atempt to gracefully shutdown

cmd/dp-filter-api/main.go
nshumoogum: Yes I think the direction from team is that main should handle all errors and be doing a lot less, e.g. use run() function to setup api server and graceful shutdown but return errors from run back to main so there is one exit point of the application. If we are heading down that route then we would want to return an error here an exit app back in main function

I will leave this up to you for now, if we do end up using `run(...) error {` then the team will need to update all applications anyway

config/config.go
```diff
@@ -53,7 +53,7 @@ func Get() (*Config, error) {
 		DatasetAPIURL:              "http://localhost:22000",
 		DatasetAPIAuthToken:        "FD0108EA-825D-411C-9B1D-41EF7727F465",
 		HealthCheckInterval:        30 * time.Second,
-		HealthCheckCriticalTimeout: 2 * time.Second,
+		HealthCheckCriticalTimeout: 90 * time.Second,
```
nshumoogum, line 56: You probably need to update config_test.go too

cmd/dp-filter-api/main.go
CarlHembrough: Makes sense - thanks for explaining 👍

api/api.go
```diff
@@ -2,18 +2,17 @@ package api
 
 import (
 	"context"
-
+	"github.com/ONSdigital/dp-api-clients-go/dataset"
+	"github.com/ONSdigital/dp-api-clients-go/middleware"
```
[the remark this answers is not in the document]
CarlHembrough: I have added the middleware function as suggested by Nathan to remove the dependency on github.com/ONSdigital/dp-api-clients-go/middleware

go.mod
nshumoogum: This has now been implemented

config/config.go
nshumoogum: Bump

PR #158, opened 2020-08-19
dependency auditing implementation

rav-pradhan:
### What

Implement dependency auditing with Nancy

### How to review

- Check changes align with other implementations for dependency auditing within our services
- Ensure correct permissions are set for the `audit.sh` file (chmod 755)

### Who can review

Anyone

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,31 +11,31 @@ LDFLAGS=-ldflags "-w -s -X 'main.Version=${VERSION}' -X 'main.BuildTime=$(BUILD_
 
 DATABASE_ADDRESS?=bolt://localhost:7687
 
-PHONY: all
+.PHONY: all
 all: audit test build
 
-PHONY: audit
+.PHONY: audit
 audit:
 	nancy go.sum
 
-PHONY: build
+.PHONY: build
 build:
 	@mkdir -p $(BUILD)/$(BIN_DIR)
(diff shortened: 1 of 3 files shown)


DavidSubiros (approved):

Makefile
```diff
@@ -11,15 +11,31 @@ LDFLAGS=-ldflags "-w -s -X 'main.Version=${VERSION}' -X 'main.BuildTime=$(BUILD_
 
 DATABASE_ADDRESS?=bolt://localhost:7687
 
+PHONY: all
```
CarlHembrough, line 14: Does `PHONY` need to be prefixed with a `.`? Most repos have the prefix, though I also noticed that the `dp-api-router` doesn't.

DavidSubiros (changes requested):
sorry, didn't see @CarlHembrough  comment before

DavidSubiros (changes requested):
Same comment as noted by @CarlHembrough in https://github.com/ONSdigital/dp-filter-api/pull/158

```Does PHONY need to be prefixed with a .? Most repos have the prefix, though I also noticed that the dp-api-router doesn't.```

Makefile
rav-pradhan: I think it does need to be prefixed with a `.` - I will update here and in `dp-api-router`, too

DavidSubiros (approved):

PR #160, opened 2020-09-03
Removed old auditing code and config

DavidSubiros:
### What

- Removed auditing (main task):
  - code, tests and dependencies
  - Kafka producer
  - Audit topic config
  - Actions constants
- Moved dependencies from go-ns to dp-net
  - Identity handlers from dp-net
  - Upgraded dependencies to versions that use dp-net ContextKey
- Introduced service package (with unit tests), to be consistent with newer APIs code structure
  - moved most of main.go code to service.go (similar to other APIs code)
  - moved http server, midelware, etc from api.go to service.go
  - identity client created only once
  - Service can be started without MongoDB

### How to review

- Unit tests should pass

### Who can review

Anyone


service/service.go
```diff
@@ -0,0 +1,302 @@
...
+	go func() {
+		log.Event(ctx, "Starting api...", log.INFO)
+		if err := svc.server.ListenAndServe(); err != nil {
+			svcErrors <- errors.Wrap(err, "failure in http listen and serve")
+		}
+	}()
+}
+
+// CreateMiddleware creates an Alice middleware chain of handlers
+func (svc *Service) createMiddleware(ctx context.Context, cfg *config.Config) alice.Chain {
```
CarlHembrough, line 160: `cfg` parameter is not used and could be removed

CarlHembrough (changes requested):
Just a small one to remove an unused parameter!

CarlHembrough (approved):

PR #167, opened 2020-10-27
Documented uninitialised mongoDB dependency case

DavidSubiros:
### What

- Added some code comments to document the case of a missing dependency (MongoDB) that does not cause the service to exit.
  - Service initialisation: do not return on error
  - Register checkers: create a critical handler

### How to review

Make sure the comments are clear.

### Who can review

Anyone

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -86,11 +86,12 @@ func (svc *Service) Init(ctx context.Context, cfg *config.Config, buildTime, git
 	svc.cfg = cfg
 
 	// Get data store.
-	// We don't want to stop this service due to a failure in connecting with mongoDB.
-	// A failing healthcheck Checker will be created in this case.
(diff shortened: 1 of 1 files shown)


redhug1:
Please can you re-arrange the comments like this:
	svc.filterStore, err = getFilterStore(svc.cfg)
	if err != nil {
		log.Event(ctx, "could not connect to mongodb", log.ERROR, log.Error(err))
		// We don't return 'err' here because we don't want to stop this service
		// due to a failure in connecting with mongoDB.
		// A failing healthcheck Checker will be created later in registerCheckers().
	}

The comment for the reason for not returning 'err' needs to be where the line of code 'return err' is normally found.

redhug1 (approved):
LGTM

PR #168, opened 2020-10-27
Refactor registerCheckers to make clearer.

redhug1:
Make the code clearer to understand.


DavidSubiros (approved):
thanks for refactoring this code. Now it's clearer. Tested: `make debug` with a nil MongoDB dependency, and validated that `\healt` shows the critical checker

PR #170, opened 2020-11-04
Implemented PATCH endpoint to patch filter dimension options.

DavidSubiros:
### What

- Implemented `PATCH /filters/{filter_blueprint_id}/dimensions/{name}` to modify dimension options
  - uses generic PATCH struct from dp-net (with basic validation, including supported op's: `add` and `remove`) - depends on https://github.com/ONSdigital/dp-net/pull/15
  - validates provided path (must be `/options/-`)
  - only starts applying patches (sequentially) if all provided patches are valid
  - if a patch operation fails the remaining operations are not attempted, and the successful patches are logged.

- Introduced MongoDB functions to add or remove multiple options
  - AddFilterDimensionOptions
  - RemoeFilterDimensionOptions

- Refactored a bit filter_dimension_options, in order to reuse the add/remove logic:
  - Add and Patch add share the same logic
  - Add, as well as patch, will fail earlier if the dimension is not found in the blueprint (the same that was done in remove) - but keeping the dimension check against dataset API if the dimension exists in the blueprint (so that if it doesn't we return UnprocessableEntity); this will prevent unnecessary calls to Dataset API.
  - Remove and Patch remove don't share the same logic, because remove returns an Error when the option is not found, whilst patch does not.
  - Extracted `getFilterBlueprint` from the logic functions, so that we call it only when necessary.

- Unit tested, making sure I did not modify any response code with the refactoring. Note: I couldn't unit-test `UnprocessableEntity` case, maybe we want to export actualCode for ErrInvalidDatasetAPIResponse in dp-api-clients.dataset?

- Modified swagger spec accordingly

### How to review

- Make sure unit tests succeed

### Who can review

Anyone


justinpjose (approved):
This all looks good. Ran the tests for it as well locally and they all pass. Nicely done David

PR #172, opened 2020-11-16
Feature/max values patch

DavidSubiros:
### What

- Limit the number of values that can be provided in a single PATCH request
  - the max applies to the sum of values among all patch operations in the array.
- Made the MongoDB add/remove queries (used by patch) more efficient
- PATCH response now returns the list of successful patch operations, instead of all option values for a dimension.
  - modified swagger spec accordingly

### How to review

- Make sure unit tests pass, and check that they cover the new feature

### Who can review

Anyone

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -30,6 +30,7 @@ type PreviewDataset interface {
 // FilterAPI manages importing filters against a dataset
 type FilterAPI struct {
 	host                 string
+	maxRequestOptions    int
 	dataStore            DataStore
 	outputQueue          OutputQueue
 	router               *mux.Router
@@ -42,6 +43,7 @@ type FilterAPI struct {
 
 // Setup manages all the routes configured to API
 func Setup(host string,
+	maxRequestOptions int,
 	router *mux.Router,
 	dataStore DataStore,
 	outputQueue OutputQueue,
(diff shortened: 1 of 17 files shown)


redhug1 (commented):
Some observations of the code:

In: api/filter_dimension_options.go
Line: 416
Missing the following at the end of the Line:  , log.ERROR, log.Error(err), logData)
similar to line 408  ?

In: api/api_test.go
Line: 18
Should the value being assigned be 1000  ?

-=-=-

I confirm that the unit tests cover the new feature (and that they work .. i adjusted new code to stop them working and stepped through new test code with debugger to confirm failure and then working when new code put back)

DavidSubiros:
> I confirm that the unit tests cover the new feature (and that they work .. i adjusted new code to stop them working and stepped through new test code with debugger to confirm failure and then working when new code put back)

Thanks for the thorough review! and for catching the log at line 416, I will fix that.

About the value of `maxRequestOptions` in api_test, I set a value of 10 for testing, in order to test the scenario of a batch PATCH  operation (in `filter_dimension_options_test.go` line 473). But if it's confusing maybe I can just set a value of 10 in that particular test?

redhug1 (approved):
Unit test works fine.
Code: LGTM

PR #173, opened 2020-11-16
Update the map so the collections are passed in as an array

HeJones:
### What

- Build the collections name for a database into an array so all the collections can be passed into the healthcheck client and checked

### How to review

Test that when either of the collections isn't in the mongodb database the healthcheck fails

### Who can review

Anyone


DavidSubiros (approved):
It looks good, nice improvement.
Checked that the `/health` endpoint still returns the expected outcome:
```
        {
            "name": "Mongo DB",
            "status": "CRITICAL",
            "message": "collection not found in database",
            "last_checked": "2020-11-16T14:47:37.210815Z",
            "last_success": null,
            "last_failure": "2020-11-16T14:47:37.210815Z"
        },
```

PR #176, opened 2020-12-04
Feature/add pagination

DavidSubiros:
### What

Implemented pagination for `GET /filters/{filter_blueprint_id}/dimensions/{name}/options` endpoint:
- Added `offset` and `limit` query parameters
- Defaults are provided in mongo config `MONGODB_LIMIT` and `MONGODB_OFFSET`
- Paginated response with `limit`, `offset`, `count` and `totalCount`
- Sort filter dimension options alphabetically and slice according to limit and offset
- Refactored API Setup function to accept the config struct (easier to maintain and incorporate new values in the future)
- Added unit tests

### How to review

- Make sure unit tests pass
- Make sure code makes sense
- You may try to run filter api locally (`make debug`) and then validate the pagination with a call to the endpoint, similar to this:
```
curl --location --request GET 'http://localhost:22100/filters/a72b7a9b-4017-4ac1-ba21-5eb5ff3f4420/dimensions/aggregate/options?offset=1&limit=3' \
--header 'X-Florence-Token:{token}'
```

### Who can review

Anyone

### DIFF
--- a/api/filter_dimension_options.go
+++ b/api/filter_dimension_options.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"net/http"
 	"net/url"
+	"sort"
 	"strconv"
 
 	"github.com/ONSdigital/dp-filter-api/models"
@@ -32,7 +33,7 @@ func (api *FilterAPI) getFilterBlueprintDimensionOptionsHandler(w http.ResponseW
 	// get limit from query parameters, or default value
 	limit, err := getPositiveIntQueryParameter(r.URL.Query(), "limit", api.defaultLimit)
 	if err != nil {
-		log.Event(ctx, "failed to obtain limit from request query paramters", log.ERROR, log.Error(err), logData)
+		log.Event(ctx, "failed to obtain limit from request query paramters", log.ERROR, logData)
 		setErrorCode(w, err)
 		return
 	}
@@ -40,7 +41,7 @@ func (api *FilterAPI) getFilterBlueprintDimensionOptionsHandler(w http.ResponseW
 	// get offset from query parameters, or default value
 	offset, err := getPositiveIntQueryParameter(r.URL.Query(), "offset", api.defaultOffset)
 	if err != nil {
-		log.Event(ctx, "failed to obtain offset from request query paramters", log.ERROR, log.Error(err), logData)
+		log.Event(ctx, "failed to obtain offset from request query paramters", log.ERROR, logData)
 		setErrorCode(w, err)
 		return
 	}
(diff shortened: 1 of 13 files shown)


api/filter_dimension_options.go
```diff
@@ -27,14 +30,30 @@ func (api *FilterAPI) getFilterBlueprintDimensionOptionsHandler(w http.ResponseW
...
+	if err != nil {
+		log.Event(ctx, "failed to obtain limit from request query paramters", log.ERROR, logData)
+		setErrorCode(w, err)
+		return
+	}
+
+	// get offset from query parameters, or default value
+	offset, err := getPositiveIntQueryParameter(r.URL.Query(), "offset", api.defaultOffset)
+	if err != nil {
+		log.Event(ctx, "failed to obtain offset from request query paramters", log.ERROR, logData)
```
CarlHembrough, line 44: typo in `parameters`
```suggestion
		log.Event(ctx, "failed to obtain offset from request query parameters", log.ERROR, logData)
```

api/filter_dimension_options.go
```diff
@@ -27,14 +30,30 @@ func (api *FilterAPI) getFilterBlueprintDimensionOptionsHandler(w http.ResponseW
 	ctx := r.Context()
 	log.Event(ctx, "get filter blueprint dimension options", log.INFO, logData)
 
+	// get limit from query parameters, or default value
+	limit, err := getPositiveIntQueryParameter(r.URL.Query(), "limit", api.defaultLimit)
+	if err != nil {
+		log.Event(ctx, "failed to obtain limit from request query paramters", log.ERROR, logData)
```
CarlHembrough, line 36: typo in `parameters`
```suggestion
		log.Event(ctx, "failed to obtain limit from request query parameters", log.ERROR, logData)
```

api/filter_dimension_options.go
```diff
@@ -61,14 +80,35 @@ func (api *FilterAPI) getFilterBlueprintDimensionOptionsHandler(w http.ResponseW
...
+			options = &models.PublicDimensionOptions{
+				Items:      []*models.PublicDimensionOption{},
+				TotalCount: len(dimension.Options),
+				Offset:     offset,
+				Limit:      limit,
+			}
+
+			// sort alphabetically and cut according to limit and offset
+			sort.Strings(dimension.Options)
+			dimension.Options = slice(dimension.Options, offset, limit)
```
CarlHembrough, line 111: don't we want to get the database to do the limit / sorting that is applied here? The database could still return a huge result set that would consume memory when it's never used.

CarlHembrough (changes requested):

api/filter_dimension_options.go
DavidSubiros: yes, I tried to do that but I think we would need to split the filters into separate filters and options collections? (similar to dataset api), or alternatively, do you know if it's possible to query parts of a document? (if it's possible we could query everything except the options, and then query only the options that we need - I haven't seen if this is possible)

CarlHembrough: It feels like we should separate the options, as there seems to be a few cases where they are not wanted. I am not sure if that it worth doing right now though. I will go ahead and approve, and we can see if we need to go further.

CarlHembrough (approved):

PR #177, opened 2020-12-04
Fix/conflict on multiple patch

DavidSubiros:
### What

- Fixed bug where multiple patch operations could sometimes cause a 409 conflict because we were using the filter blueprint timestamp from the first time we read it. Now we read it before applying each patch.
- Fixed bug in remove options, where we were not removing options if any of the provided options did not exist. Now we verify exactly what we need to remove.

### How to review

- Make sure unit tests pass
- (feel free to ask me for a quick demo) it can be tested by running filter api with `make debug`, and sending the following patch operations to `http://localhost:22100/filters/{filterID}/dimensions/aggregate`:
  - sending patch with multiple operations returns 200OK :
```
[
    {
        "op": "add",
        "path": "/options/-",
        "value": ["cpih1dim1G110200"]
    },
    {
        "op": "remove",
        "path": "/options/-",
        "value": ["cpih1dim1G110200"]
    }
]
```
  - sending remove patch with existing and inexistent items results in the existing items being removed:
```
[
    {
        "op": "remove",
        "path": "/options/-",
        "value": ["cpih1dim1G110200", "inexistent"]
    }
]
```
### Who can review

Anyone

### DIFF
--- a/api/filter_dimension_options.go
+++ b/api/filter_dimension_options.go
@@ -349,7 +349,7 @@ func (api *FilterAPI) removeFilterBlueprintDimensionOption(ctx context.Context, 
 	return nil
 }
 
-// removeFilterBlueprintDimensionOption removes all provided options, if any option did not exist, it will be ignored
+// removeFilterBlueprintDimensionOption removes all provided options.
 func (api *FilterAPI) removeFilterBlueprintDimensionOptions(ctx context.Context, filterBlueprint *models.Filter, dimensionName string, options []string, logData log.Data) error {
 
 	// check if any option has been provided
@@ -357,30 +357,32 @@ func (api *FilterAPI) removeFilterBlueprintDimensionOptions(ctx context.Context,
 		return nil
 	}
 
-	// Check if provided dimension and options exists
-	hasDimension, hasOptions, missingOptions := findDimensionAndOptions(filterBlueprint, dimensionName, options)
-
+	// Check if provided dimension and options exists in filter blueprint
+	hasDimension, hasAllOptions, missingOptions := findDimensionAndOptions(filterBlueprint, dimensionName, options)
 	if !hasDimension {
 		return filters.ErrDimensionNotFound
 	}
 
-	if !hasOptions {
+	// find options that actually need to be removed according to the existing options before applying any change
+	optionsToRemove := []string{}
+	if hasAllOptions {
+		optionsToRemove = options
+	} else {
+		for _, option := range options {
+			if _, found := missingOptions[option]; !found {
+				optionsToRemove = append(optionsToRemove, option)
+			}
+		}
+	}
+
+	// if none of the provided options were present, we don't need to remove anything
+	if len(optionsToRemove) == 0 {
 		log.Event(ctx, "options do not exist in the dimension, nothing to remove", log.INFO, log.Data{})
 		return nil
 	}
 
-	timestamp := filterBlueprint.UniqueTimestamp
-
-	// find options that need to be removed (i.e. are present in filer blueprint)
-	optionsToRemove := []string{}
-	for _, option := range options {
-		if _, found := missingOptions[option]; !found {
-			optionsToRemove = append(optionsToRemove, option)
-		}
-	}
-
 	// remove necessary options from DB
-	return api.dataStore.RemoveFilterDimensionOptions(filterBlueprint.FilterID, dimensionName, optionsToRemove, timestamp)
+	return api.dataStore.RemoveFilterDimensionOptions(filterBlueprint.FilterID, dimensionName, optionsToRemove, filterBlueprint.UniqueTimestamp)
 }
 
 // Handler for a list of patch operations against the dimension options
--- a/api/filter_dimension_options.go
+++ b/api/filter_dimension_options.go
@@ -212,16 +212,8 @@ func (api *FilterAPI) addFilterBlueprintDimensionOptionHandler(w http.ResponseWr
 	}
 	ctx := r.Context()
 
-	// request filterBlueprint before updating it
-	filterBlueprint, err := api.getFilterBlueprint(ctx, filterBlueprintID)
-	if err != nil {
-		log.Event(ctx, "error getting filter blueprint dimension option", log.ERROR, log.Error(err), logData)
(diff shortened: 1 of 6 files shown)


api/filter_dimension_options.go
```diff
@@ -457,20 +451,31 @@ func (api *FilterAPI) patchFilterBlueprintDimensionHandler(w http.ResponseWriter
...
-	if !hasDimension {
-		return successful, filters.ErrDimensionNotFound
-	}
+	dimensionValidated := false
 
 	// apply patch operations sequentially, stop processing if one patch fails, and return a list of successful patches operations
 	for _, patch := range patches {
 		options := removeDuplicateAndEmptyOptions(patch.Value)
+
+		// Get filter Blueprint in all iterations to prevent conflicts due to the timestamp check in mongoDB when adding or removing values
```
daiLlew, line 464: This is a good point. I think for all updates we should always be updating the latest stored version.

api/filter_dimension_options_test.go
```diff
@@ -397,6 +397,45 @@ func TestSuccessfulPatchFilterBlueprintDimension(t *testing.T) {
...
+		reader := strings.NewReader(`[
+			{"op":"remove", "path": "/options/-", "value": ["33", "123456789"]}
+		]`)
+		r, err := http.NewRequest("PATCH", "http://localhost:22100/filters/12345678/dimensions/age", reader)
+		So(err, ShouldBeNil)
+
+		w := httptest.NewRecorder()
+		api := Setup(host, maxRequestOptions, mux.NewRouter(), &mocks.DataStore{}, &mocks.FilterJob{}, &mocks.DatasetAPI{}, previewMock, enablePrivateEndpoints, downloadServiceURL, downloadServiceToken, serviceAuthToken)
+		api.router.ServeHTTP(w, r)
+		So(w.Code, ShouldEqual, http.StatusOK)
```
daiLlew, line 410: Should this test also be asserting the expected remove was carried out?

api/filter_dimension_options_test.go
```diff
@@ -397,6 +397,45 @@ func TestSuccessfulPatchFilterBlueprintDimension(t *testing.T) {
...
+			{"op":"remove", "path": "/options/-", "value": []}
+		]`)
+		r, err := http.NewRequest("PATCH", "http://localhost:22100/filters/12345678/dimensions/age", reader)
+		So(err, ShouldBeNil)
+
+		w := httptest.NewRecorder()
+		api := Setup(host, maxRequestOptions, mux.NewRouter(), &mocks.DataStore{}, &mocks.FilterJob{}, &mocks.DatasetAPI{}, previewMock, enablePrivateEndpoints, downloadServiceURL, downloadServiceToken, serviceAuthToken)
+		api.router.ServeHTTP(w, r)
+		So(w.Code, ShouldEqual, http.StatusOK)
+	})
```
daiLlew, line 437: Same for each of these - they verify the request was successful and returned the expected HTTP status but they aren't asserting the correct update was carried out. 
(Sorry to be a pain 😁 )

api/filter_dimension_options.go
```diff
@@ -349,38 +349,40 @@ func (api *FilterAPI) removeFilterBlueprintDimensionOption(ctx context.Context,
...
+		for _, option := range options {
+			if _, found := missingOptions[option]; !found {
+				optionsToRemove = append(optionsToRemove, option)
+			}
 		}
 	}
 
+	// if none of the provided options were present, we don't need to remove anything
+	if len(optionsToRemove) == 0 {
+		log.Event(ctx, "options do not exist in the dimension, nothing to remove", log.INFO, log.Data{})
```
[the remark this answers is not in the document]
DavidSubiros: thanks, good catch!

api/filter_dimension_options_test.go
DavidSubiros: Yes, I found that the mocks are not create through `moq`, they are simply an implementation of the interface, covering different cases. I started to try to make the mocks extensions of the `moq` mock, but it was quite a lot of changes. I can continue that work, it will also be useful for other tests.

api/filter_dimension_options.go
```diff
@@ -349,38 +349,40 @@ func (api *FilterAPI) removeFilterBlueprintDimensionOption(ctx context.Context,
...
 	// check if any option has been provided
 	if len(options) == 0 {
 		return nil
 	}
 
-	// Check if provided dimension and options exists
-	hasDimension, hasOptions, missingOptions := findDimensionAndOptions(filterBlueprint, dimensionName, options)
-
+	// Check if provided dimension and options exists in filter blueprint
+	hasDimension, hasAllOptions, missingOptions := findDimensionAndOptions(filterBlueprint, dimensionName, options)
```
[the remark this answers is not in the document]
DavidSubiros: You're right, this function assumes that the provided model is the latest one. At the moment this is only called by the patch endpoint and provides always the latest model. I can leave a comment, or maybe move the call to `getFilterBlueprint` inside `addFilterBlueprintDimensionOptions` and `removeFilterBlueprintDimensionOptions`, as we need to do it for each patch anyway.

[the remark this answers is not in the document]
DavidSubiros: I decided to move the 'getFilterBlueprint' inside the functions that need to modify it, as you mentioned.

mocks/datastore.go
```diff
@@ -30,25 +31,130 @@ type DataStore struct {
...
+		UpdateFilterFunc:                 ds.UpdateFilter,
+		UpdateFilterOutputFunc:           ds.UpdateFilterOutput,
+		AddEventToFilterOutputFunc:       ds.AddEventToFilterOutput,
+	}
+	ds.Mock = mock
+	return ds
+}
+
+// NotFound sets NotFound flag to true
+func (ds *DataStore) NotFound() *DataStore {
```
daiLlew, line 68: I think this might not be **thread safe** - if several tests use the same mock in parallel it could end up with funky / inconsistent test results. I think the solution would be to add a lock around each field update.

**However**, that it a lot of additional work that so maybe for now just state this in the struct Godoc? If anyone runs into that issue at a later time they can fix when its required. Feels like a suitable compremise.

daiLlew (approved):

PR #178, opened 2020-12-09
Validate dimension options with paginated calls

DavidSubiros:
### What

Use paginated calls (offset and limit) in order to validate dimension option against dataset api. We need to do this in order to successfully filter big datasets like census.

### How to review

- Make sure unit tests pass
- (info only): tested locally with cpih dataset and a batch size of 2.

### Who can review

Anyone


go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/ONSdigital/dp-filter-api
 go 1.15
 
 require (
-	github.com/ONSdigital/dp-api-clients-go v1.30.0
+	github.com/ONSdigital/dp-api-clients-go v1.30.5-0.20201209084153-6c2e12c0444e
```
rav-pradhan, line 6: Update this to latest release of the library

rav-pradhan (commented):
Couple small comments, but otherwise looks fine (talked through with David)

rav-pradhan (approved):

PR #179, opened 2020-12-14
updated to use latest dp-kafka version

andre-urbani:
### What

Updated to use latest dp-kafka version

### How to review

Confirm changes are correct

### Who can review

Anyone


go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/ONSdigital/dp-api-clients-go v1.31.0
 	github.com/ONSdigital/dp-graph/v2 v2.3.0
 	github.com/ONSdigital/dp-healthcheck v1.0.5
-	github.com/ONSdigital/dp-kafka v1.1.7
+	github.com/ONSdigital/dp-kafka/v2 v2.1.1
```
DavidSubiros, line 9: can we use `dp-kafka v2.1.2` please? (I removed a log that was too noisy and made debugging of issues more difficult)

service/service.go
```diff
@@ -56,10 +56,17 @@ var getObservationStore = func(ctx context.Context) (observationStore *graph.DB,
...
 
+var pConfig *kafka.ProducerConfig
+
+var cfg *config.Config
+
 // getProducer returns a kafka producer
-var getProducer = func(ctx context.Context, kafkaBrokers []string, topic string, envMax int) (kafkaProducer kafka.IProducer, err error) {
+var getProducer = func(ctx context.Context, kafkaBrokers []string, topic string) (kafkaProducer kafka.IProducer, err error) {
+	pConfig := &kafka.ProducerConfig{
+		KafkaVersion: &cfg.KafkaVersion,
```
DavidSubiros, line 66: We should provide MaxMessageBytes to producer config, as it is provided by the general config:
```
		MaxMessageBytes: &cfg.KafkaMaxBytes,
```

service/service.go
```diff
@@ -56,10 +56,17 @@ var getObservationStore = func(ctx context.Context) (observationStore *graph.DB,
 	return observationStore, errorConsumer, nil
 }
 
+var pConfig *kafka.ProducerConfig
```
DavidSubiros, line 59: I think this variable is not needed, you already define `pConfig` inside getProducer (I think it's the right place)

service/service.go
```diff
@@ -56,10 +56,17 @@ var getObservationStore = func(ctx context.Context) (observationStore *graph.DB,
...
+
+var cfg *config.Config
```
DavidSubiros, line 61: This config is not initialised, so when it's used by getProducer it will be using default values only. Can you remove this variable and provide config as parameter accepted by `getProducuer` please? 
Then when you call getProducer from Init(), you can provide the initialised config (`svc.cfg`)

service/service.go
```diff
@@ -56,10 +56,17 @@ var getObservationStore = func(ctx context.Context) (observationStore *graph.DB,
...
+
 // getProducer returns a kafka producer
-var getProducer = func(ctx context.Context, kafkaBrokers []string, topic string, envMax int) (kafkaProducer kafka.IProducer, err error) {
+var getProducer = func(ctx context.Context, kafkaBrokers []string, topic string) (kafkaProducer kafka.IProducer, err error) {
```
DavidSubiros, line 64: `getProducer` should accept a config, which is then used to create pConfig.
maybe the signature can be:
`func(ctx context.Context, cfg *config.Config, topic string)`
and then you can obtain the values for kafkaVersion (as you do now), as well as kafkaBrokers and maxBytes

DavidSubiros (commented):
A couple of comments, the most important is about passing an initialised config, please let me know if you have any question.

DavidSubiros (approved):

PR #180, opened 2020-12-16
Removed code related to old /healthcheck endpoint

andre-urbani:
### What

Removed code related to old /healthcheck endpoint, as per this ticket - https://trello.com/c/ECS9esvf/4743-filter-api-remove-transitional-health-check-code-1d

### How to review

Run service locally and confirm that /healthcheck endpoint returns a 404 error

### Who can review

Anyone


DavidSubiros (approved):
I would have suggested removing the `else if  ... NotFound() ...` but then I realised that the Trello ticket required that the return code should be 404, and by default, it's 401, so I think the change is correct.

Also validated that the ci is already using go version 1.15, approving.

PR #188, opened 2021-01-19
Feature/etag

DavidSubiros:
CAUTION! Breaking change. Will need to merge after:
- https://github.com/ONSdigital/dp-frontend-filter-dataset-controller/pull/262
- https://github.com/ONSdigital/dp-frontend-dataset-controller/pull/195

### What

Implemented `ETag` logic for the filter blueprint API endpoints:
- ETag is a hash of the filter document (marshalled according to the `bson` tags)
  - Used sha1 for efficiency (it is very fast but not cryptographically safe. For this use case it is ok)
- 'Reader' endpoints don't check any ETag value but return the current value as a response `ETag` header
- 'Writer' endpoints check the provided ETag value, which must be provided as a value in a compulsory `If-Match` header
  - If the header is not provided, a 400 Bad Request response is returned
  - If the provided value doesn't match the existing one, a 409 Conflict is returned
  - If the provided value is an asterisk, then the resource will not be validated and the update will go ahead

At the mongoDB level:
- Updated filter select queries to include the ETag value, if provided
- ETag is not enforced. If not provided, it will not be used in the select queries

Unit tests:
- The dataset mock has been modified in order to check and return ETags in a similar way as described above. We keep track of the number of ETag updates and then validate or update accordingly.

### How to review

- Make sure unit test changes make sense, cover the new functionality, and pass
- (info only - optional) I tested the calls with Postman, validating returned ETags, providing '*', correct and incorrect If-Match values
 
### Who can review

Anyone


redhug1 (approved):
Looks OK

PR #193, opened 2021-02-08
Feature/pagination update

DavidSubiros:
### What

Updated paginated endpoints to comply with the latest API standards defined in https://github.com/ONSdigital/dp/blob/master/standards/API_STANDARDS.md :
- limit=0 returns an empty list of items instead of all items
- default limit = 20
- cap maximum limit to 1000
- negative values for offset and limit result in a 400 Bad Request response
- Updated swagger spec

### How to review

- Make sure changes comply with the new standards
- Make sure changes make sense
- Make sure unit tests pass

### Who can review

Anyone


api/api_test.go
```diff
@@ -44,8 +44,9 @@ func cfg() *config.Config {
 		ServiceAuthToken:         serviceAuthToken,
 		EnablePrivateEndpoints:   enablePrivateEndpoints,
 		MaxDatasetOptions:        200,
+		DefaultMaxLimit:          1000,
 		MongoConfig: config.MongoConfig{
-			Limit:  0,
+			Limit:  20,
```
daiLlew, line 49: Just so I can understand this change - what does this limit do?

api/filter_dimension_options.go
```diff
@@ -28,22 +26,39 @@ func (api *FilterAPI) getFilterBlueprintDimensionOptionsHandler(w http.ResponseW
...
+		logData["limit"] = limitParameter
+		limit, err = validatePositiveInt(limitParameter)
+		if err != nil {
+			log.Event(ctx, "failed to obtain limit from request query parameters", log.ERROR, logData)
+			setErrorCode(w, err)
+			return
+		}
+	}
+
+	if limit > api.maxLimit {
```
daiLlew, line 59: Is there value in putting the max limit in the log data? Seems like it would be useful for anyone investigating an issue in this area.

api/filter_dimensions.go
```diff
@@ -22,20 +22,39 @@ func (api *FilterAPI) getFilterBlueprintDimensionsHandler(w http.ResponseWriter,
...
+		limit, err = validatePositiveInt(limitParameter)
+		if err != nil {
+			log.Event(ctx, "failed to obtain limit from request query parameters", log.ERROR, logData)
+			setErrorCode(w, err)
+			return
+		}
+	}
+
+	if limit > api.maxLimit {
+		err = filters.ErrInvalidQueryParameter
```
daiLlew, line 56: Same here for the max limit.

api/filter_dimensions.go
```diff
@@ -390,7 +409,7 @@ func (api *FilterAPI) checkNewFilterDimensionOptions(ctx context.Context, dimens
 // createPublicDimensions wraps createPublicDimension for converting arrays of dimensions
 func createPublicDimensions(inputDimensions []models.Dimension, host, filterID string) []*models.PublicDimension {
 
-	var outputDimensions []*models.PublicDimension
+	var outputDimensions = []*models.PublicDimension{}
```
daiLlew, line 412:
````
outputDimensions := make([]*models.PublicDimension, 0)
````

daiLlew (commented):
A few minor questions but otherwise LGTM.

api/api_test.go
DavidSubiros: This value is the default limit for pagination.
When we do a GET request to an endpoint that can return a list of items, we may only return a subset of the total items. The caller can provide query paramters:
- limit: the maximum number of items that they want to request
- offset: the first index that they want to requrest
if they don't provide a limit, a value  of 20 will be used.
(well, in fact this is the value for testing, but it's the same in the real config)

api/filter_dimension_options.go
DavidSubiros: good point, I will put it

daiLlew (approved):

PR #195, opened 2021-02-24
Filter API issues

Marianasilvago:
### What
document state field for FilterOutputResponse
add time format in swagger for event time.
refactor linkobject fields to pointer type to ensure omitempty is honoured

### How to review
check the code changes.

### Who can review
Anyone

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -747,6 +747,8 @@ definitions:
       time:
         type: string
         description: "The time of the event happened"
+        example: "2016-07-17T08:38:25.316+0000"
+        format: string
       type:
         type: string
         description: "The type of event which happened"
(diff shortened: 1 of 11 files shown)


nshumoogum (changes requested):

swagger.yaml
```diff
@@ -536,17 +536,36 @@ definitions:
...
+            type: string
+      - type: object
+        properties:
+          links:
+            allOf:
+              - $ref: '#/definitions/FilterLinks'
+              - type: object
+                properties:
+                  filter_blueprint:
+                    allOf:
```
nshumoogum, line 557: No need to include `allOf` here as there is only one object to inherit for a `filter_blueprint`

nshumoogum (changes requested):

nshumoogum (approved):

PR #196, opened 2021-02-26
introduce id field for filterOutput and filter blueprint responses.

Marianasilvago:
### What

FilterOutputResponse and BluePrint do not contain a overall id field. Copied over filter_id to id field.

### How to review

check the code changes.

### Who can review

Anyone with golang understanding.

### DIFF
--- a/models/models.go
+++ b/models/models.go
@@ -45,6 +45,7 @@ type Filter struct {
 	LastUpdated     time.Time           `bson:"last_updated"               json:"-"`
 	ETag            string              `bson:"e_tag"                      json:"-"`
 
+	ID         string      `bson:"-"                    json:"id,omitempty"`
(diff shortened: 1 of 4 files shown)


jondewijones (approved):

models/models.go
```diff
@@ -45,6 +45,7 @@ type Filter struct {
 	LastUpdated     time.Time           `bson:"last_updated"               json:"-"`
 	ETag            string              `bson:"e_tag"                      json:"-"`
 
+	ID         string      `bson:"-"                    json:"id,omitempty"`
```
sandypadmanabhan, line 48: Should this be `bson:"id"?

Marianasilvago: We don't want to persist this id to the Data store. The filter_id is already getting persisted.
We only want to expose it to keep it in parity with other APIs.

PR #204, opened 2021-05-20
Add more event types to filterOutputs and allow better updating

eldeal:
### What

- Move event consts to the models package
- Move all event logic into separate file in models package
- Add validation for events
- Refactor add events to ensure multiple events can be submitted on a PUT /filter-outputs/<id> request

### How to review

Ensure new filter events can be submitted to the API (potentially using the /filter-outputs/<id>/events POST endpoint)
Ensure multiple filter events can be submitted on the same PUT request

### Who can review

anyone


models/event.go
```diff
@@ -0,0 +1,56 @@
...
+	EventFilterOutputQueryStart   = "FilterOutputQueryStart"
+	EventFilterOutputQueryEnd     = "FilterOutputQueryEnd"
+	EventFilterOutputCSVGenStart  = "FilterOutputCSVGenStart"
+	EventFilterOutputCSVGenEnd    = "FilterOutputCSVGenEnd"
+	EventFilterOutputXLSXGenStart = "FilterOutputXLSXGenStart"
+	EventFilterOutputXLSXGenEnd   = "FilterOutputXLSXGenEnd"
+	EventFilterOutputCompleted    = "FilterOutputCompleted"
+)
+
+type Event struct {
```
daiLlew, line 20: Public type should have Godoc.

daiLlew (dismissed):

api/filter_outputs.go
```diff
@@ -325,10 +315,37 @@ func (api *FilterAPI) addEventHandler(w http.ResponseWriter, r *http.Request) {
...
+	if len(events) > 0 {
+		for _, e := range events {
+			if err := e.Validate(); err != nil {
+				log.Error(ctx, "filter output event failed validation", err, logData)
+				return filters.NewBadRequestErr(err.Error())
+			}
+
+			if err := api.dataStore.AddEventToFilterOutput(ctx, id, e); err != nil {
+				log.Error(ctx, "failed to add event to filter output", err, logData)
+				return err
```
rafahop, line 330: Should we be logging errors and then returning them? We end up with the same error logged multiple times?

api/filter_outputs.go
```diff
@@ -136,19 +136,9 @@ func (api *FilterAPI) updateFilterOutput(ctx context.Context, filterOutputID str
...
-		completedEvent := &models.Event{
-			Type: EventFilterOutputCompleted,
-			Time: time.Now(),
-		}
-
-		if err = api.dataStore.AddEventToFilterOutput(ctx, filterOutput.FilterID, completedEvent); err != nil {
-			log.Error(ctx, "failed to add event to filter output", err, logData)
-			return err
-		}
+	if err = api.addEvents(ctx, filterOutput.Events, filterOutputID, isNowStatusCompleted); err != nil {
```
rafahop, line 139: Unsure if this change is right. I believe that by calling `addEvents` we will be doing the same as before (adding the completed event) AND also adding all `filterOutput.Events` to the filter output (which wasn't happening previously). Is this intentional? If so, maybe we should reflect it in the tests?

api/filter_outputs.go
eldeal: Good point, have removed

api/filter_outputs.go
eldeal: It's a good point. I think I was trying to get too fancy where there was no need. I've changed things a bit so there's a helper function for the FilterOutputUpdates to use to make sure we're not adding duplicate events. And then for now I've left the /events POST endpoint as it is - assuming someone intentionally adding a single event will be less likely to inadvertently spam us the same event multiple times.

Added some unit tests for the helper funcs as well

models/event_test.go
```diff
@@ -0,0 +1,159 @@
...
+					{
+						Type: "Event2",
+						Time: time.Now(),
+					},
+				},
+			}
+
+			Convey("Then the new event should be kept", func() {
+				new.RemoveDuplicateEvents(current)
+				So(len(new.Events), ShouldEqual, 1)
```
rafahop, line 81: I would also assert that the event that remains is the right one if possible (ie Event2, and not Event1). It would probably make more sense if we had filters with multiple events though

rafahop (approved):

models/event.go
```diff
@@ -0,0 +1,81 @@
...
+	events := []*Event{}
+
+	for _, e := range filter.Events {
+		found := false
+		for _, ce := range currentFilter.Events {
+			//compare the values not the pointers
+			if *e == *ce {
+				found = true
+			}
+		}
```
rafahop, line 40: We could stop iterating the inner for loop once we found the event
```suggestion
		for _, ce := range currentFilter.Events {
			//compare the values not the pointers
			if *e == *ce {
				found = true
				break
			}
		}
```

Or slightly change the for loop if we prefer not to add `break` statements:

```suggestion
		for i :=0; !found && i < len(currentFilterEvents.Events; i++ {
		        ce := currentFilterEvents.Events[i]
			//compare the values not the pointers
			if *e == *ce {
				found = true
			}
		}
```

rafahop (approved):

PR #208, opened 2021-07-07
Feature/remove preview

redhug1:
### What

Remove preview endpoint from Documentation.
Remove the preview endpoint from the filter-api routes.
Remove the preview handler.
Applied go generate (using latest moq v0.2.2) and fix ensuing import cycles.

### How to review

Look over what code was removed
Run tests
Run Audit

### Who can review

Anyone familiar with filter-api


api/filter_dimension_options.go
```diff
@@ -498,7 +498,15 @@ func (api *FilterAPI) patchFilterBlueprintDimensionHandler(w http.ResponseWriter
 			http.Error(w, err.Error(), http.StatusBadRequest)
 			return
 		}
-		totalValues += len(patch.Value)
+		v2, ok := patch.Value.([]interface{})
```
DavidSubiros, line 501: I think here we should call `getStringArrayFromInterface` in order to guarantee that the values are correct, and `getStringArrayFromInterface` should return an error if any value is incorrect.
Otherwise, we might start applying patches with some invalid input.
```
v2, err := getStringArrayFromInterface(patch.Value)
if err != nil {
...
```

api/filter_dimensions.go
```diff
@@ -438,18 +438,37 @@ func createPublicDimension(dimension models.Dimension, host, filterID string) *m
...
+			if v3 != "" {
+				result = append(result, v3)
 			}
 		}
 	}
 
 	return result
 }
+
+func getStringArrayFromInterface(elements interface{}) []string {
```
DavidSubiros, line 457: I would suggest that this function returns ([]string, error), returning an error in any `!ok` case.

api/filter_dimension_options.go
```diff
@@ -541,7 +549,8 @@ func (api *FilterAPI) patchFilterBlueprintDimension(ctx context.Context, filterB
 
 	// apply patch operations sequentially, stop processing if one patch fails, and return a list of successful patches operations
 	for _, patch := range patches {
-		options := removeDuplicateAndEmptyOptions(patch.Value)
+		allOptions := getStringArrayFromInterface(patch.Value)
```
DavidSubiros, line 552: if `getStringArrayFromInterface` returns an error, we should check it here and abort (it should not happen because this should have already failed before, during validation):
```
allOptions, err := getStringArrayFromInterface(patch.Value)
if err != nil {
    return successful, eTag, err
}
```

DavidSubiros (commented):
Just a couple of small comments, which were unrelated to the main task. I also tried `make debug` and validated that the graph dependency doesn't appear any more in the `/health` endpoint

api/filter_dimension_options.go
redhug1: Thanks for review.
Improvement applied.

api/filter_dimensions.go
redhug1: Thanks for review.
Improvement applied.

api/filter_dimension_options.go
redhug1: Thanks for review.
Improvement applied.

DavidSubiros (approved):

PR #214, opened 2021-08-13
add new dp-kafka version and config,docs for using TLS

gedge:
### What

make this app kafka-over-TLS ready

### How to review

kafka TLS is feature-flagged off by default, so ensure no substantive changes

### Who can review

!gedge


config/config.go
```diff
@@ -70,24 +76,23 @@ func Get() (*Config, error) {
...
+			BindAddr:                   "localhost:27017",
+			Database:                   "filters",
+			FiltersCollection:          "filters",
+			OutputsCollection:          "filterOutputs",
+			Limit:                      20, // Default limit for mongoDB queries that do not provide an explicit limit
+			Offset:                     0,  // Default offset for mongoDB queries that do not provide an explicit offset
+			Username:                   "",
+			Password:                   "",
+			IsSSL:                      false,
+			EnableStrongReadConcern:    false,
```
bpathak-ons, line 88: I think these values should be configured as env variables same as kafka-config.

gedge: You are right. However:

1. my editor ran `go fmt` on this file (because I changed other lines), so these changes were only in **formatting** (alignment of values) of existing code
    - my changes are in the lighter green (on my dark-mode view) and highlight whitespace only changes
   
2. https://github.com/ONSdigital/dp-filter-api/blob/develop/config/config.go#L42 indicates that they are indeed configured by env vars (the section you see on my PR is just setting defaults, the link is where those defaults are overwritten by env vars)

so I hope I can safely ignore this comment

bpathak-ons (approved):
looks ok

PR #227, opened 2022-02-18
Feature/add assert filter type middleware

ChrysmOre:
### What

Added assert.FilterType middleware to intercept requests on flexible type filters to be fowarded to `dp-cantabular-filter-flex-api`

Only added to GET /filter/id path for now, as the equivalent endpoints are developed in `dp-cantabular-filter-flex-api` we'll need to add this middleware to those routes too

### How to review

Check changes make sense, tests pass

To test manually:
- Bring up Cantabular Import journey and checkout to this branch
- make POST request to /filters with a cantabular_flexible_filter type dataset to create filter
- Test request is forwarded to dp-filter-flexapi when making GET request to /filters/{id} (will throw an error if not, also check the logs for the `dp-cantabular-filter-flex-api` service to be sure

### Who can review

Describe who worked on the changes, so that other people can review.


api/filters_test.go
```diff
@@ -989,6 +989,76 @@ func TestRequestForwardingMiddleware(t *testing.T) {
...
+
+			r, err := http.NewRequest("GET", "http://localhost:22100/filters/foo", nil)
+			So(err, ShouldBeNil)
+			filterApi.Router.ServeHTTP(w, r)
+
+			Convey("A call to datastore is made to check the filter type", func() {
+				So(len(datastoreMock.GetFilterCalls()), ShouldEqual, 2)
+			})
+
+			Convey("The request is forwarded to dp-cantabular-filter-flex-api", func() {
```
goofballLogic, line 1058:
```suggestion
			Convey("The request is NOT forwarded to dp-cantabular-filter-flex-api", func() {
```

api/filters_test.go
```diff
@@ -989,6 +989,76 @@ func TestRequestForwardingMiddleware(t *testing.T) {
...
+				So(len(filterFlexAPIMock.ForwardRequestCalls()), ShouldEqual, 0)
```
goofballLogic, line 1059: where should it go instead? Do we need to test for that?

api/filters_test.go
```diff
@@ -1032,5 +1102,42 @@ func TestRequestForwardingMiddleware(t *testing.T) {
...
+
+			r, err := http.NewRequest("GET", "http://localhost:22100/filters/foo", nil)
+			So(err, ShouldBeNil)
+			filterApi.Router.ServeHTTP(w, r)
+
+			Convey("A call to dataset-api is not made", func() {
+				So(len(datastoreMock.GetFilterCalls()), ShouldEqual, 0)
+			})
+
+			Convey("The request is forwarded to dp-cantabular-filter-flex-api", func() {
```
goofballLogic, line 1138:
```suggestion
			Convey("The request is NOT forwarded to dp-cantabular-filter-flex-api", func() {
```

middleware/assert.go
```diff
@@ -112,10 +80,69 @@ func (a *Assert) DatasetType(next http.Handler) http.Handler {
...
+				})
+			}
+			return
+		}
+
+		next.ServeHTTP(w, r)
+	})
+}
+
+func (a *Assert) FilterType(next http.Handler) http.Handler {
```
goofballLogic, line 95: where are the tests for this logic?

service/interfaces.go
```diff
@@ -28,7 +31,20 @@ type HealthChecker interface {
...
 // MongoDB defines the required methods from MongoDB package
 type MongoDB interface {
-	api.DataStore
+	AddFilter(ctx context.Context, filter *models.Filter) (*models.Filter, error)
+	GetFilter(ctx context.Context, filterID, eTagSelector string) (*models.Filter, error)
+	UpdateFilter(ctx context.Context, updatedFilter *models.Filter, timestamp primitive.Timestamp, eTagSelector string, currentFilter *models.Filter) (newETag string, err error)
+	GetFilterDimension(ctx context.Context, filterID string, name, eTagSelector string) (dimension *models.Dimension, err error)
+	AddFilterDimension(ctx context.Context, filterID, name string, options []string, dimensions []models.Dimension, timestamp primitive.Timestamp, eTagSelector string, currentFilter *models.Filter) (newETag string, err error)
+	RemoveFilterDimension(ctx context.Context, filterID, name string, timestamp primitive.Timestamp, eTagSelector string, currentFilter *models.Filter) (newETag string, err error)
+	AddFilterDimensionOption(ctx context.Context, filterID, name, option string, timestamp primitive.Timestamp, eTagSelector string, currentFilter *models.Filter) (newETag string, err error)
```
goofballLogic, line 40: can't find anywhere `AddFilterDimensionOption` is used but...

goofballLogic (commented):
Finished comments

api/filters_test.go
goofballLogic: this still appears to be a problem

PR #228, opened 2022-02-25
add filterflex api url to config

ChrysmOre:
### What

Added FilterFlexapiURL to config instead of being hardcoded

### How to review

Check change makes sense

### Who can review

Anyone


redhug1 (approved):
LGTM

PR #230, opened 2022-03-03
fix middleware error messages

ChrysmOre:
### What

Fixed up error messages in assert middleware to make them less misleading

### How to review

Check change makes sense

### Who can review

Anyone


PauloPortugal (approved):

goofballLogic (approved):
fmt.Sprintf("lgtm")

PR #232, opened 2022-03-08
updated swagger spec to match target contract

goofballLogic:
### What

Attempted to correct the filter-api swagger spec for POST to /filters

1. Added "population_type" field (type string) which is an optional field understood by upstream service dp-cantabular-filter-flex-api
2. Added missing "is_based_on" field do Dataset
3. Added the model for IsBasedOn - copied directly from dp-cantabular-filter-flex

### How to review

Things to consider:
1. population_type field is using the OpenAPI `type` keyword to set the data type, but dp-cantabular-filter-flex service uses the `format` keyword for the same thing - not sure if it matters that they're different (official spec doesn't seem to clarify, but normally base types use the `type` keyword)
2. Adding the is_based_on field to the model for Dataset will add this as an optional field for _every endpoint which uses this definition_. Is that ok?

### Who can review

@ChrysmOre and @franmoore05

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -551,6 +551,10 @@ definitions:
           description: "A list of dimensions in the filter job"
           items:
              $ref: '#/definitions/DimensionOptions'
+        population_type:
+          description: "If specifying a custom filter, the population type that the data should be queried from"
+          example: "Teaching-Dataset"
+          type: string
   UpdateFilterRequest:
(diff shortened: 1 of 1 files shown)


ChrysmOre (approved):

PauloPortugal (approved):
Cool!!!

We should move this swagger spec into the endpoints and document it there and have the swagger spec be auto-generated.

PR #234, opened 2022-03-29
update swagger spec

ChrysmOre:
### What

Updated swagger spec to match actual behaviour throughout

### How to review

check changes make sense

### Who can review

anyone


api/api.go
```diff
@@ -90,10 +90,12 @@ func Setup(
 	)
 
 	// routes
-	api.Router.Handle("/filters", assert.DatasetType(http.HandlerFunc(api.postFilterBlueprintHandler))).Methods("POST")
+	//api.Router.Handle("/filters", assert.DatasetType(http.HandlerFunc(api.postFilterBlueprintHandler))).Methods("POST")
```
jayfreestone, line 93: Probably should remove.

api/filters.go
```diff
@@ -38,14 +39,19 @@ const (
...
 	submitted := r.FormValue("submitted")
 	logData := log.Data{"submitted": submitted}
 	ctx := r.Context()
 	log.Info(ctx, "create filter blueprint", logData)
 
+	b, err := io.ReadAll(r.Body)
+	if err != nil {
+		return
+	}
+	fmt.Println("DEBUG: body:", string(b))
```
jayfreestone, line 53: Looks like a dev log line crept in.

middleware/assert.go
```diff
@@ -41,10 +41,10 @@ func NewAssert(r responder, d datasetAPIClient, f filterFlexAPIClient, ds datast
 
 func (a *Assert) DatasetType(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if !a.enabled {
+		//if !a.enabled {
 			next.ServeHTTP(w, r)
 			return
-		}
+		//}
```
jayfreestone, line 47: Probably worth just removing the conditional rather than commenting it out.

jayfreestone (approved):
👍

PauloPortugal (approved):
We should move away from manually updating swagger files and rely on [Go-Swagger](https://goswagger.io/) to generate this for us.

This way we are closer to the implementation.

PR #235, opened 2022-03-31
Feature/filter routing middleware

saminahbab:
### What

Added middleware logic to 4 routes as per defined in ticket.
   - GET request to  /filters/{filter_blueprint_id}
   - PUT request to  /filters/{filter_blueprint_id}
   - GET request to  /filters/{filter_blueprint_id}
   - PUT request to  /filter-outputs/{filter_output_id}
   
Added route based id parsing method for dataset based forwarding.
Added and ammended tests to for verifying middleware logic.

Created Middleware Mock generation method for cleaner test setup.

### How to review

Run `make test` for unit tests

### Who can review

Anyone from team b


middleware/assert.go
```diff
@@ -39,6 +39,48 @@ func NewAssert(r responder, d datasetAPIClient, f filterFlexAPIClient, ds datast
...
+		if !a.enabled {
+			next.ServeHTTP(w, r)
+			return
+		}
+
+		vars := mux.Vars(r)
+		filterID := vars["filter_blueprint_id"]
+
+		ctx := r.Context()
+		if !a.enabled {
```
ChrysmOre, line 55: repeated block?

api/filters_test.go
```diff
@@ -1169,6 +1274,7 @@ func TestRequestForwardingMiddleware(t *testing.T) {
 			})
 
 			Convey("The request is not forwarded to dp-cantabular-filter-flex-api", func() {
+				// This should make it however
```
ChrysmOre, line 1277: Not sure what this comment means?

api/filters_test.go
```diff
@@ -1129,6 +1172,68 @@ func TestRequestForwardingMiddleware(t *testing.T) {
...
+			Convey("A call is not made to datastore to check the filter type", func() {
+				So(len(datastoreMock.GetFilterCalls()), ShouldEqual, 0)
+			})
+			Convey("The request is forwarded to dp-cantabular-filter-flex-api", func() {
+				So(len(filterFlexMock.ForwardRequestCalls()), ShouldEqual, 1)
+			})
+		})
+
+		Convey("When there is a GET request to /filters/{filter_blueprint_id}/dimensions and the dataset type is cantabular", func() {
+			filterFlexMock, datasetAPIMock, datastoreMock := mock.GenerateMocksForMiddleware(200, cantabularFlexibleTable, 1, "not-applicable")
```
ChrysmOre, line 1199: What's this 200? If it's a status code it's better to use the const `http.StatusOK` for clarity

mock/middlewareMock.go
```diff
@@ -0,0 +1,47 @@
+package mock
```
ChrysmOre, line 1: In terms of the file name, it's generally the convention to use snake_case for filenames. Furthermore it's unnecessary to repeat yourself in the package + filename.. I.E `mock/middleware.go` should suffice

api/api.go
```diff
@@ -91,10 +91,10 @@ func Setup(
 
 	// routes
 	api.Router.Handle("/filters", assert.DatasetType(http.HandlerFunc(api.postFilterBlueprintHandler))).Methods("POST")
-	api.Router.Handle("/filters/{filter_blueprint_id}", assert.FilterType(http.HandlerFunc(api.getFilterBlueprintHandler))).Methods("GET")
-	api.Router.Handle("/filters/{filter_blueprint_id}/dimensions", assert.FilterType(http.HandlerFunc(api.getFilterBlueprintDimensionsHandler))).Methods("GET")
+	api.Router.Handle("/filters/{filter_blueprint_id}", assert.DatasetTypeFromRoute(assert.FilterType(http.HandlerFunc(api.getFilterBlueprintHandler)))).Methods("GET")
```
ChrysmOre, line 94: What's going on here? Why do we need two levels of middleware?

ChrysmOre, line 94: Not sure how you can assert the dataset type from a filter ID?

middleware/assert.go
```diff
@@ -39,6 +39,48 @@ func NewAssert(r responder, d datasetAPIClient, f filterFlexAPIClient, ds datast
...
+			next.ServeHTTP(w, r)
+			return
+		}
+
+		// Get the dataset
+		dataset, err := a.DatasetAPI.Get(ctx, "", a.svcAuthToken, "", filterID)
```
ChrysmOre, line 61: How can you get the dataset from a filter ID?

ChrysmOre (commented):
Some comments + not sure if I understand what's going on properly, but looks like you're trying to call GetDataset with a filter ID, which won't work?

middleware/assert.go
saminahbab: This has been removed.

api/api.go
saminahbab: As per offline conversations surrounding the ticket, the question has been addressed with the logic changed.

mock/middlewareMock.go
saminahbab: Changed.

middleware/assert.go
```diff
@@ -39,6 +39,42 @@ func NewAssert(r responder, d datasetAPIClient, f filterFlexAPIClient, ds datast
...
+		filterOutput, err := a.store.GetFilterOutput(ctx, filterOutputID)
+		if err != nil {
+			a.respond.Error(ctx, w, statusCode(err), er{
+				err: errors.Wrap(err, "failed to get filter output"),
+				msg: fmt.Sprintf("failed to get filter output"),
+			})
+			return
+		}
+
+		if filterOutput.Type == flexible {
```
jackrehaag, line 64: Currently I think that the `type` of `filterOutput` is not populated in the datastore. There is some work pending to correct this but for the time being it might be worth using the `filterOutput.Links.FilterBlueprint.ID` to find the `filter` with the correct `type`.

api/filters_test.go
```diff
@@ -1180,13 +1203,15 @@ func TestRequestForwardingMiddleware(t *testing.T) {
 					Dataset: &models.Dataset{
 						Version: 1,
 					},
+					Type: "not-flexible",
```
jackrehaag, line 1206: Test name indicates a flexible filter type but the mock returns Type: "not-flexible"

api/filters_test.go
```diff
@@ -1315,39 +1362,59 @@ func TestRequestForwardingMiddleware(t *testing.T) {
...
 			So(err, ShouldBeNil)
+
 			filterApi.Router.ServeHTTP(w, r)
 
-			Convey("A call to dataset-api is not made", func() {
-				So(len(datastoreMock.GetFilterCalls()), ShouldEqual, 0)
+			Convey("A call is not made to datastore to check the filter type", func() {
+				So(len(datastoreMock.GetFilterOutputCalls()), ShouldEqual, 0)
+			})
+			Convey("The request is not  forwarded to dp-cantabular-filter-flex-api", func() {
```
jackrehaag, line 1379: Nitpick (not blocking), double space.

jackrehaag (commented):
I think it might be worth adding some tests in `assert_test.go` for the new `FilterOutputFilterType` function that tests the calls to the datastore mocks in a bit more detail, this would ensure that the correct IDs are being passed through to `GetFilterOutput`. 
At the moment I suspect the tests are not passing the correct vars via the request as `mux.setVars` is not being used.

middleware/assert.go
saminahbab: you are right, it is actually not, having just checked. 
Although using the `filterOutput.Links.FilterBlueprint.ID` to find the filter would then result in two database calls, which feels really hacky, not sure if that is a great way to go. Would rather just say that this work is finished but blocked from releasing until the work pending to correct is finished

api/filters_test.go
saminahbab: If you expand the bar closest above to this comment, you'll see that it relates to the appropriate  `convey` statement

saminahbab:
> I think it might be worth adding some tests in `assert_test.go` for the new `FilterOutputFilterType` function that tests the calls to the datastore mocks in a bit more detail, this would ensure that the correct IDs are being passed through to `GetFilterOutput`. At the moment I suspect the tests are not passing the correct vars via the request as `mux.setVars` is not being used.

@jackrehaag the appropriate test in `assert_test.go` has been added with changes to the `mux.SetURLVars` call in other methods, ensuring that the correct IDs are being used in a database call

middleware/assert.go
jackrehaag: @saminahbab Sounds reasonable but probably best to check with @franmoore05 and co to make sure.

middleware/assert_test.go
```diff
@@ -124,7 +124,54 @@ func TestAssertFilterType(t *testing.T){
...
+				return &models.Filter{
+					Type: "foobar",
+					Dataset: &models.Dataset{
+						Version: 2,
+					},
+					ID: filterID,
+				}, nil
+			}
+
+			Convey("When an incoming request pases through the asert.FilterOutputType middleware", func() {
```
jackrehaag, line 139: Would it be possible to add a test for when the filter output is not flexible but is still checked by `assert.FilterOutputFilterType` and therefore not forwarded to the filter-flex-api?

saminahbab: @jackrehaag line 1231 of `api/filter_test.go` should do what you seek.

jackrehaag: I think it would be best to have both cases covered in `assert_test.go` that way we're covered if the filter in question is modified or deleted.

saminahbab: Np. Added tests that cover both cases, with the appropriate conditions in `assert_test.go`. I will keep this conversation resolved for that reason

jackrehaag (approved):

PR #239, opened 2022-04-19
Add submit POST endpoint for filters

jackrehaag:
### What

Adds a `POST` `/filters/filter_blueprint_id/submit` endpoint that returns an error if the filter is not flexible and forwards the request to the filter flex API if it is.

### Who can review

Anyone.


api/filters.go
```diff
@@ -209,6 +210,21 @@ func (api *FilterAPI) getFilterBlueprintHandler(w http.ResponseWriter, r *http.R
...
 
+func (api *FilterAPI) postFilterBlueprintSubmitHandler(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	filterBlueprintID := vars["filter_blueprint_id"]
+
+	logData := log.Data{
+		"filter_blueprint_id": filterBlueprintID,
+	}
+	ctx := r.Context()
+	log.Info(ctx, "put filter blueprint", logData)
```
rafahop, line 221: Isn't this a post request?

api/filters_test.go
```diff
@@ -138,6 +138,82 @@ func TestSuccessfulAddFilterBlueprint_PublishedDataset(t *testing.T) {
...
+			})
+
+			Convey("Then an error is returned", func() {
+				So(w.Code, ShouldEqual, http.StatusBadRequest)
+			})
+		})
+	})
+}
+
+func TestUnSuccessfulPostFilterBlueprint_NonFlexible(t *testing.T) {
```
rafahop, line 177: I would have a single function to test this new functionality and use the conveys for the different scenarios? Maybe we could reuse some common setup that way?

jackrehaag: I initially had the same thought but wanted to keep the current style in the test file of using separate functions.

PR #242, opened 2022-04-22
WIP: Clean up proxying

jayfreestone:
### What

Describe what you have changed and why.

### How to review

Describe the steps required to test the changes.

### Who can review

Describe who worked on the changes, so that other people can review.

### DIFF
--- a/middleware/assert_test.go
+++ b/middleware/assert_test.go
@@ -23,6 +23,7 @@ import (
 
 const (
 	filterBlueprintID = "filter_blueprint_id"
+	filterOutputID    = "filter_output_id"
 	testBody          = "test body"
 	testToken         = "testToken"
 )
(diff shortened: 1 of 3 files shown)


middleware/assert_test.go
```diff
@@ -14,106 +14,149 @@ import (
...
-				Convey("The response should have body, status and headers as returned by dp-filter-flex-api", func(){
+				Convey("The response should have body, status and headers as returned by dp-filter-flex-api", func() {
 					So(len(datastoreMock.GetFilterCalls()), ShouldEqual, 1)
+					So(datastoreMock.GetFilterCalls()[0].FilterID, ShouldEqual, testID)
 					So(len(filterFlexAPIMock.ForwardRequestCalls()), ShouldEqual, 1)
-					So(w.Code, ShouldEqual, expectedResponse.StatusCode)
+					So(w.Code, ShouldEqual, http.StatusOK)
 					So(w.HeaderMap.Get("X-Test"), ShouldResemble, "Value")
 					So(w.HeaderMap.Get("X-Foo"), ShouldResemble, "")
 					So(w.Body.String(), ShouldResemble, testBody)
```
jayfreestone, line 137: You might be able to abstract some of these into shared assertions if they're consistent:

```go
func theRequestShouldBeForwarded(actual interface{}, _ ...interface{}) string {
	w := actual.(*httptest.ResponseRecorder)
	headers := w.Header()

	if headers.Get("X-Foo") != "" {
		return "Test header for X-Foo was found"
	}

	if headers.Get("X-Test") != "Value" {
		return "Test header for X-Test was found"
	}

	if w.Body.String() != testBody {
		return "Test body was found instead of test handler response"
	}

	return ""
}

// So(w, theRequestShouldBeForwarded)
```
Maybe could even move some of these string values into constants, idk.

middleware/assert_test.go
```diff
@@ -14,106 +14,149 @@ import (
...
+			filterIsFlexible := true
+			Convey("When an incoming request passes through the assert.FilterType middleware", func() {
+				assertIsEnabled := true
+				datastoreMock, filterFlexAPIMock, _, assert := assertTestSetup(filterIsFlexible, datasetIsFlexible, assertIsEnabled, testBluePrintID)
 
 				w := httptest.NewRecorder()
 
 				testID := "12345678"
-				r, err := http.NewRequest("GET", "http://localhost:1234/test/" + testID, nil)
+				r, err := http.NewRequest("GET", "http://localhost:1234/test/"+testID, nil)
```
jayfreestone, line 119: Might be worth just getting rid of this, since it has no effect on the test. The vars are just injected by `SetURLVars`:

```go
r, err := http.NewRequest(http.MethodGet, "/", nil)
So(err, ShouldBeNil)
```

middleware/assert_test.go
jayfreestone: If you don't go this route, still worth getting rid of `HeaderMap` as it's deprecated tho.

middleware/assert_test.go
```diff
@@ -221,25 +325,20 @@ func TestAssertDatasetType(t *testing.T){
...
-					responder.New(),
-					datasetAPIMock,
-					filterFlexAPIMock,
-					&mock.DataStore{},
-					testToken,
-					false,
-				)
+			Convey("When an incoming request passes through a disabled assert.FilterType middleware", func() {
+				assertIsEnabled := false
+				datasetIsFlexible := true
```
jayfreestone, line 340: These shadow vars in the same scope. I would probably just move the others into the block they're needed, or maybe just inline the bools at the callsite?

middleware/assert_test.go
```diff
@@ -14,106 +14,149 @@ import (
...
+		datastoreMock,
+		testToken,
+		assertIsEnabled,
+	)
+	return datastoreMock, filterFlexAPIMock, datasetAPIMock, assert
+}
+
+func TestAssertFilterType(t *testing.T) {
+	Convey("Given a healthy dp-cantabular-filter-flex-api", t, func() {
+		testBluePrintID := ""
```
jayfreestone, line 108: Isn't the test blueprint ID the same as the filter ID? i.e. `testID`?

middleware/assert_test.go
```diff
@@ -14,106 +14,149 @@ import (
...
-			Convey("When an incoming request passes through the assert.FilterType middleware", func(){
-				assert := middleware.NewAssert(
-					responder.New(),
-					&mock.DatasetAPI{},
-					filterFlexAPIMock,
-					datastoreMock,
-					testToken,
-					true,
-				)
+func assertTestSetup(filterIsFlexible, datasetIsFlexible, assertIsEnabled bool, testBluePrintID string) (*apimock.DataStoreMock, *apimock.FilterFlexAPIMock, *apimock.DatasetAPIMock, *middleware.Assert) {
```
jayfreestone, line 90: 👍

Might even be able to take this further and have it run the req, although doesn't really save that many lines:

```go
func runassertTest(req *http.Request, filterIsFlexible, datasetIsFlexible, assertIsEnabled bool, testBluePrintID string) (*apimock.DataStoreMock, *apimock.FilterFlexAPIMock, *apimock.DatasetAPIMock, *httptest.ResponseRecorder) {
	filterFlexAPIMock := setupFilterFlexAPIMock()
	datastoreMock := setupDataStoreMock(filterIsFlexible, testBluePrintID)
	datasetAPIMock := setupDatasetAPIMock(datasetIsFlexible)

	assert := middleware.NewAssert(
		responder.New(),
		datasetAPIMock,
		filterFlexAPIMock,
		datastoreMock,
		testToken,
		assertIsEnabled,
	)

	w := httptest.NewRecorder()
	next := http.HandlerFunc(testHandler)
	f := assert.FilterType(next)
	f.ServeHTTP(w, req)

	return datastoreMock, filterFlexAPIMock, datasetAPIMock, w
}
```

middleware/assert_test.go
jayfreestone: Also any time you use `NewRequest` here you can use the `httptest` version, which panics if there is an error:

```go
r := httptest.NewRequest(http.MethodGet, "/", nil)
```

middleware/assert_test.go
jayfreestone: Actually ignore me, because the method will be different.

middleware/assert_test.go
```diff
@@ -14,106 +14,149 @@ import (
...
+	datasetAPIMock.GetFunc = func(ctx context.Context, ut, st, cid, dsid string) (dataset.DatasetDetails, error) {
+		dataset := dataset.DatasetDetails{
+			Type: "not_flexible",
 		}
+		if datasetIsFlexible {
+			dataset.Type = "cantabular_flexible_table"
+		}
+		return dataset, nil
+	}
+	return datasetAPIMock
```
jayfreestone, line 87: Can just define these inline if you like:

```go
func setupDatasetAPIMock(datasetIsFlexible bool) *apimock.DatasetAPIMock {
	return &apimock.DatasetAPIMock{
		GetFunc: func(_ context.Context, _, _, _, _ string) (dataset.DatasetDetails, error) {
			details := dataset.DatasetDetails{
				Type: "not_flexible",
			}
			if datasetIsFlexible {
				details.Type = "cantabular_flexible_table"
			}
			return details, nil
		},
	}
}
```

No matter either way.

middleware/assert_test.go
PauloPortugal: > If you don't go this route, still worth getting rid of HeaderMap as it's deprecated tho

This is true, but the `httptest` package still uses it.

PauloPortugal: > If you don't go this route, still worth getting rid of HeaderMap as it's deprecated tho

This is true, but the `httptest` package still uses it.

PauloPortugal: I will keep the code as it is, as there are different expected values to be tested.

middleware/assert_test.go
PauloPortugal: Addressed.

middleware/assert_test.go
PauloPortugal: Addressed.

middleware/assert_test.go
PauloPortugal: Addressed.

middleware/assert_test.go
PauloPortugal: Addressed.

PR #248, opened 2022-05-10
Search filterOutput collection via "filter_id" or "id"

PauloPortugal:
### What

In CMD we search by `filter_id`, but on Cantabular we search by a new
field called `id`.

This addresses the approach of simply changing the MongoDB query.

Resolves: [5690](https://trello.com/c/ZDZgbMyz/5690-copy-getfilteroutput-function-from-dp-filter-api-and-amend-for-cantabular-filter-output-search)

### How to review

Review the changes and run them locally

### Who can review

Any ONS developer

### DIFF
--- a/models/models.go
+++ b/models/models.go
@@ -47,7 +47,7 @@ type Filter struct {
 	LastUpdated     time.Time           `bson:"last_updated"               json:"-"`
 	ETag            string              `bson:"e_tag"                      json:"-"`
 
-	ID         string      `bson:"-"                    json:"id,omitempty"`
+	ID         string      `bson:"id"                    json:"id,omitempty"`
 	Dataset    *Dataset    `bson:"dataset"              json:"dataset"`
 	InstanceID string      `bson:"instance_id"          json:"instance_id"`
 	Dimensions []Dimension `bson:"dimensions,omitempty" json:"dimensions,omitempty"`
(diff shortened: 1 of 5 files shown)


models/models.go
```diff
@@ -47,7 +47,7 @@ type Filter struct {
 	LastUpdated     time.Time           `bson:"last_updated"               json:"-"`
 	ETag            string              `bson:"e_tag"                      json:"-"`
 
-	ID         string      `bson:"-"                    json:"id,omitempty"`
```
saminahbab, line 50: this is probably fine, as long as we let the stakeholders beyond team b know that we are going this far

mongo/filterstore.go
```diff
@@ -377,9 +377,9 @@ func (s *FilterStore) CreateFilterOutput(ctx context.Context, filter *models.Fil
 
 // GetFilterOutput returns a filter output resource
 func (s *FilterStore) GetFilterOutput(ctx context.Context, filterID string) (*models.Filter, error) {
-	// NOTE: previously was filter_id.
-	// This should be _id/id. Confirm during PR.
-	query := bson.M{"filter_id": filterID}
+	// we have to match either the filter_id (in CMD) or the id (in Cantabular)
```
saminahbab, line 380: I think a slightly longer comment explicitly saying why it led to having to support in this way would be helpful for future context

j-s-rawat (approved):
LGTM

models/models.go
PauloPortugal: @franmoore05 @eldeal :point_up_2:

PR #250, opened 2022-05-10
Search filterOutput collection via "filter_id" or "id"

PauloPortugal:
### What

In CMD we search by `filter_id`, but on Cantabular we search by a new
field called `id`.

Resolves: [5690](https://trello.com/c/ZDZgbMyz/5690-copy-getfilteroutput-function-from-dp-filter-api-and-amend-for-cantabular-filter-output-search)

### How to review

Review code.

### Who can review

Any develop

### DIFF
--- a/api/datastore.go
+++ b/api/datastore.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"context"
+
 	"go.mongodb.org/mongo-driver/bson/primitive"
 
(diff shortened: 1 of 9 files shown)


PauloPortugal:
This was a small spike/implementation/approach that we have discussed alongside https://github.com/ONSdigital/dp-filter-api/pull/248. 
We have decided not to go with this approach.

PR #252, opened 2022-05-17
Feature/2.18.0

PauloPortugal:
### What

7dd8778 Search filterOutput collection via "filter_id" or "id" (5 days ago) @PauloPortugal 
0db8ec2  Update MongoDB library and README (7 days ago) @PatrickConnollyONS 
### How to review

Review changes

### Who can review

People above and other devs


rafahop (changes requested):
Unsure why the commits here are different to the ones in develop? They have the same content but are different commits. Hasn't `feature/2.18.0` been cut from `develop`?
Also ideally should be prefixed `release`, not `feature`

PR #265, opened 2022-07-01
feat: add delete options proxy

saminahbab:
Trello ticket #5735

### What
DELETE "/filters/{id}/dimensions/{name}/options"
add to routes to proxy over to cantabular filter flex
api and dummy handler to represent bad state

### How to review

View the changes.
It was judged that testing was not required as middleware tests are already detailed in testing that requests are appropriately forwarded.

### Who can review

Team B

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -99,9 +99,9 @@ func Setup(
 	api.Router.Handle("/filters/{filter_blueprint_id}/dimensions/{name}", assert.FilterType(http.HandlerFunc(api.putFilterBlueprintDimensionHandler))).Methods("PUT")
 	api.Router.Handle("/filters/{filter_blueprint_id}/dimensions/{name}", assert.FilterType(http.HandlerFunc(api.getFilterBlueprintDimensionHandler))).Methods("GET")
 	api.Router.Handle("/filters/{filter_blueprint_id}/dimensions/{name}/options/{option}", assert.FilterType(http.HandlerFunc(api.removeFilterBlueprintDimensionOptionHandler))).Methods("DELETE")
-
 	api.Router.Handle("/filters/{filter_blueprint_id}/dimensions/{name}/options/{option}", assert.FilterType(http.HandlerFunc(api.addFilterBlueprintDimensionOptionHandler))).Methods("POST")
 	api.Router.Handle("/filters/{filter_blueprint_id}/dimensions/{name}/options", assert.FilterType(http.HandlerFunc(api.getFilterBlueprintDimensionOptionsHandler))).Methods("GET")
+	api.Router.Handle("/filters/{filter_blueprint_id}/dimensions/{name}/options", assert.FilterType(http.HandlerFunc(api.filterFlexNullEndpointHandler))).Methods("DELETE")
 
 	api.Router.Handle("/filter-outputs/{filter_output_id}", assert.FilterOutputType(http.HandlerFunc(api.getFilterOutputHandler))).Methods("GET")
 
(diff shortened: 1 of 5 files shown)


api/api.go
```diff
@@ -99,9 +99,9 @@ func Setup(
 	api.Router.Handle("/filters/{filter_blueprint_id}/dimensions/{name}", assert.FilterType(http.HandlerFunc(api.putFilterBlueprintDimensionHandler))).Methods("PUT")
 	api.Router.Handle("/filters/{filter_blueprint_id}/dimensions/{name}", assert.FilterType(http.HandlerFunc(api.getFilterBlueprintDimensionHandler))).Methods("GET")
 	api.Router.Handle("/filters/{filter_blueprint_id}/dimensions/{name}/options/{option}", assert.FilterType(http.HandlerFunc(api.removeFilterBlueprintDimensionOptionHandler))).Methods("DELETE")
-
 	api.Router.Handle("/filters/{filter_blueprint_id}/dimensions/{name}/options/{option}", assert.FilterType(http.HandlerFunc(api.addFilterBlueprintDimensionOptionHandler))).Methods("POST")
 	api.Router.Handle("/filters/{filter_blueprint_id}/dimensions/{name}/options", assert.FilterType(http.HandlerFunc(api.getFilterBlueprintDimensionOptionsHandler))).Methods("GET")
+	api.Router.Handle("/filters/{filter_blueprint_id}/dimensions/{name}/options", assert.FilterType(http.HandlerFunc(api.dummyHandler))).Methods("DELETE")
```
PauloPortugal, line 105: I am not a fan of these dummy handlers, but I understand **the ticket is about it**.
I would rather call it a really proper package/function name and keep the dummy behaviour.

saminahbab: Sorry Paulo I'm not really sure what you mean.
Do you mean, just a handler that looks the same, but is just in its own package?

api/dummy.go
```diff
@@ -0,0 +1,29 @@
...
+func (api *FilterAPI) dummyHandler(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	filterBlueprintID := vars["filter_blueprint_id"]
+
+	logData := log.Data{
+		"filter_blueprint_id": filterBlueprintID,
+	}
+
+	ctx := r.Context()
+	log.Error(ctx, "Should not hit this endpoint", errors.New("unhandled state"), logData)
```
ChrysmOre, line 27: No blank line at the end of file.. Also it does actually need to handle a request if it does end up here by accident. There is an example of another endpoit The logged error message is also bad, it won't make any sense when seen in the logs (there's no mention of what endpoint it is, for one). Also you should keep the language more formal ('invalid request destination' or something). Look at `api.postFilterBlueprintSubmitHandler` for a good example.

api/api.go
saminahbab: Handler has been renamed to its own file with a more appropriate name, I will leave you to check and resolve

api/dummy.go
saminahbab: The endpoint now returns a bad request, with logging of the route taken, and more formal language. I will leave you to resolve

api/filter_dimension_options_test.go
```diff
@@ -1545,6 +1545,47 @@ func TestGetFilterBlueprintDimensionOptionsFilterTypeAssertion(t *testing.T) {
 	})
 }
 
+func TestDummyHandler(t *testing.T) {
+	// SAMIN
```
PauloPortugal, line 1549: I would remove this comment :)

PauloPortugal (approved):
Looks good to me.
I would remove the `// SAMIN` comment.
Not a blocker.

Happy to approve again if you want to address that comment.

api/filter_dimension_options_test.go
saminahbab: Oh yeah forgot about that!

PauloPortugal (approved):

PR #274, opened 2022-10-14
UpdateFilterOutput controlled with feature flag

j-s-rawat:
### What
US: https://trello.com/c/wGtu6FIe/5889-require-permission-to-access-put-filter-outputs

For supporting the web journey, functionality of updateFilterOutput is controlled by feature flag. On web domain, if the feature flag is on and the endpoint is called with right access token, updateFilteroutput journey will progress similar publishing domain.
 
Related to: https://github.com/ONSdigital/dp-cantabular-filter-flex-api/pull/136

### How to review

 All test are :green_circle:

### Who can review

Anyone

PR #276, opened 2022-10-18
Feature/5895 put filters name

PauloPortugal:
### What

 * The `PUT /filters/{filter-id}/dimensions/{dimension-name}` should  now only be accessible by the web frontend services.
 * Our API users should no longer be able to access this endpoint.
 * This endpoint should be made private, with a similar solution  to what we have implemented for the `PUT /filter-outputs` endpoint  to ensure only authenticated services can access the functionality.

Resolves: [5895](https://trello.com/c/hsYufI1J/5895-make-the-put-filters-filter-id-dimensions-dimension-name-only-accessible-from-the-website)

### How to review

Sense check it and ensure tests are :green_circle:

### Who can review

Any ONS developer


j-s-rawat (approved):
LGTM, might need rebase once https://github.com/ONSdigital/dp-filter-api/pull/274/files is release

PR #283, opened 2022-10-25
updated api-clients-go

thomasridd:
### What

Update api-clients-go

There was nothing needed except updating the version number on this one which makes me very suspicious. Please give plenty of though as to where I might have gone wrong

### How to review

Check tests
Check code

### Who can review

Go dev


franmoore05 (approved):
I've tested this locally as much as I can and it resolves the topic errors that I saw in sandbox this morning so I think it's good to go.

I manually manipulated the data in my local mongodb to add the topic fields to complete the test.

PR #286, opened 2022-11-03
add lowest_geography to dataset model

andre-urbani:
### What

Add lowest_geogrpahy to Dataset model
Trello - https://trello.com/c/lmWyoTS6/5923-add-lowestgeography-to-filterdataset-model-in-filter-api-and-filter-flex-api

### How to review

Confirm changes match those requested in Trello ticket

### Who can review

Anyone


ChrysmOre (commented):
Why are these changes needed for filter-api... lowest_geography isn't part of cmd surely? And wouldn't this break any cmd filters as there wouldn't be a lowest_geography sent?

andre-urbani:
Changes not required. See comments above

PR #287, opened 2022-11-25
Feature/add transactions

PatrickConnollyONS:
### What

Refactor the patchFilterBlueprintDimensionHandler to use the new transactional interface in the dp-mongodb library

### How to review

Review code and ensure all test are passing

### Who can review

Anyone

### DIFF
--- a/api/filter_dimension_options.go
+++ b/api/filter_dimension_options.go
@@ -517,58 +517,52 @@ func (api *FilterAPI) patchFilterBlueprintDimensionHandler(w http.ResponseWriter
 	}
 
 	// apply the patches to the filter blueprint dimension options
-	successfulPatches, newETag, err := api.patchFilterBlueprintDimension(ctx, filterBlueprintID, dimensionName, patches, logData, eTag)
-	if err != nil {
-		logData["successful_patches"] = successfulPatches
-		log.Error(ctx, "error patching filter blueprint dimension options", err, logData)
-		setErrorCodeFromError(w, err)
-		if len(successfulPatches) > 0 {
-			if err := WriteJSONBody(ctx, successfulPatches, w, logData); err != nil {
-				log.Error(ctx, "error writing JSON body during filter blueprint patch error handling", err, logData)
+	var newETag interface{}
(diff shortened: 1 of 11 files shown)


api/filter_dimension_options.go
```diff
@@ -517,23 +517,46 @@ func (api *FilterAPI) patchFilterBlueprintDimensionHandler(w http.ResponseWriter
...
-			}
-		}
 		return
 	}
 
 	// set content type, marshal and write response
 	setJSONPatchContentType(w)
-	setETag(w, newETag)
-	if err := WriteJSONBody(ctx, successfulPatches, w, logData); err != nil {
+	setETag(w, newETag.(string))
```
ChrysmOre, line 558: Probably best not to presume the interface will cast correctly here as it will panic if it doesn't. To check the underlying value is a string you can do something like:

```
	if s, ok := newEtag.(string); ok {
		setEtag(s)
	} else {
		// handle error/default etc
	}
```

api/filter_dimension_options.go
```diff
@@ -517,23 +517,46 @@ func (api *FilterAPI) patchFilterBlueprintDimensionHandler(w http.ResponseWriter
...
+		// apply patch operations sequentially, stop processing if one patch fails, and return a list of successful patches operations
+		for _, patch := range patches {
+			allOptions, err = getStringArrayFromInterface(patch.Value)
+			if err != nil {
+				return tag, err
+			}
+			options = RemoveDuplicateAndEmptyOptions(allOptions)
+
+			if patch.Op == dprequest.OpAdd.String() {
+				tag, err = api.addFilterBlueprintDimensionOptions(transactionCtx, filterBlueprintID, dimensionName, options, logData, tag)
```
ChrysmOre, line 536: Small thing but as Go style guides suggest keeping variable names short and transaction variables are often referred to as `tx` maybe `txCtx` would be a suitable name for the context

api/filter_dimension_options.go
PatrickConnollyONS: yeah was relying on the fact that it is going to be a valid string if no error from RunTransaction, but I will add the guard

ChrysmOre (approved):

PR #296, opened 2023-04-05
Stop logging 'options' as its giving logstash a bad day

redhug1:
### What

It might also be causing Out Of Memory restarts in nomad.
Update libs after dependabot notification.

### How to review

Inspect cod.

### Who can review

Anyone


franmoore05 (approved):
LGTM

PR #306, opened 2024-01-03
Added Kafka Otel instrumentation

gordonkerrONS:
What
Added Kafka Otel instrumentation


api/api.go
```diff
@@ -31,7 +31,7 @@ type FilterFlexAPI interface {
 
 // OutputQueue - An interface used to queue filter outputs
 type OutputQueue interface {
-	Queue(output *models.Filter) error
+	Queue(ctx context.Context, output *models.Filter) error
```
jessjenkins, line 34: Technically this is a breaking change to [a publicly released API](https://pkg.go.dev/github.com/ONSdigital/dp-filter-api@v0.0.0-20230731091332-87f5b26600c3/api#OutputQueue) consider extending the module instead if you can or else releasing as a V2 of the module (we use a branching v2 strategy rather than a dir based one if you do this).

filterOutputQueue/filterOutputQueue.go
```diff
@@ -1,15 +1,17 @@
...
 
 	"github.com/ONSdigital/dp-filter-api/models"
 	"github.com/ONSdigital/dp-filter-api/schema"
+	kafka "github.com/ONSdigital/dp-kafka/v4"
 )
 
 // Output is an object containng the filter output queue channel
 type Output struct {
-	FilterOutputQueue chan []byte
+	FilterOutputQueue chan kafka.BytesMessage
```
jessjenkins, line 14: Breaking change

api/api.go
gordonkerrONS: Ah ok, so you mean release a new major version of the dp-filter-api? Looks like we are currently on v2.29.0 so can release v3.0.0

